PR #3, opened 2024-03-25
First commit bringing GeoTiff elevation loader

rsutormin:


usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
...
+            raster cells
+    :return: 3-size tuple (header, data, CRS of the raster)
+    """
+
+    with rasterio.open(file_path) as src:
+        transform = src.transform
+        ll_corner = transform * (0, src.height)
+        input_nodata = src.nodata
+        elv_header = {
+            "ncols": int(src.width),
```
waltaskew, line 22: It looks like these are already ints and floats, so we shouldn't need to call the int & float functions here.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
+import rasterio.features
+
+
+def load_elevation_from_geotiff(
+    file_path: str, load_data: bool = False, result_nodata: float = None
+):
+    """
+    Loading elevation raster data from GeoTIFF file.
+    :param file_path: path to a file to load from
+    :param load_data: indicates that in addition to header, data should also be
```
waltaskew, line 10: I'd expect a function named 'load_elevation_from_geotiff' to load the elevation data by default. I'd either
- have a separate 'read_tiff_header' function for callers to use if they only want the header
- name the parameter 'header_only' and have the default be false

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
...
+                result_nodata if result_nodata is not None else input_nodata
+            ),
+        }
+        elv_data = None
+
+        if load_data:
+            data = src.read(1)
+            if result_nodata is not None:
+                data[data == input_nodata] = result_nodata
+            elv_data = data.tolist()
```
waltaskew, line 37: Why return this as a list rather than a Numpy matrix?

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
...
+
+def load_elevation_from_geotiff(
+    file_path: str, load_data: bool = False, result_nodata: float = None
+):
+    """
+    Loading elevation raster data from GeoTIFF file.
+    :param file_path: path to a file to load from
+    :param load_data: indicates that in addition to header, data should also be
+            loaded
+    :param result_nodata: optional new value for special case of no data in
```
waltaskew, line 12: I'd name & describe this as something like

:param no_data_value: Value to set in the returned data to indicate absence of data. If not supplied, the no-data value defined in the tiff file itself will be used.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
+import rasterio.features
+
+
+def load_elevation_from_geotiff(
+    file_path: str, load_data: bool = False, result_nodata: float = None
+):
+    """
+    Loading elevation raster data from GeoTIFF file.
+    :param file_path: path to a file to load from
```
waltaskew, line 9: Let's capitalize and add punctuation to these ala
https://numpydoc.readthedocs.io/en/latest/format.html#parameters

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
+import rasterio.features
+
+
+def load_elevation_from_geotiff(
+    file_path: str, load_data: bool = False, result_nodata: float = None
+):
+    """
+    Loading elevation raster data from GeoTIFF file.
```
waltaskew, line 8: Per
https://numpydoc.readthedocs.io/en/latest/format.html#short-summary

this should be
```
"""Load elevation raster data from GeoTIFF file.

...
```

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
...
+            "yllcorner": float(ll_corner[1]),
+            "cellsize": float(transform[0]),
+            "nodata_value": float(
+                result_nodata if result_nodata is not None else input_nodata
+            ),
+        }
+        elv_data = None
+
+        if load_data:
+            data = src.read(1)
```
waltaskew, line 34: I think the layer or layers read is something the caller would want to specify.

usl_pipeline/tests/formats/test_geotiff_loader.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+from usl_pipeline.formats import geotiff_loader
+from rasterio import CRS
+
+
+class GeodiffLoaderTest(unittest.TestCase):
```
waltaskew, line 6: We can use the pytest style without the need for unnittest subclasses ala
https://docs.pytest.org/en/8.0.x/#a-quick-example

usl_pipeline/tests/formats/test_geotiff_loader.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+from usl_pipeline.formats import geotiff_loader
```
waltaskew, line 2: per
https://peps.python.org/pep-0008/#imports

```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
```

so the usl_pipeline import should come last. We'd also prefer to import modules to keep the namespaces intact, so` import rasterio` rather than `from rasterio import CRS`. So I'd have here

```
import unittest

import rasterio

from usl_pipeline.formats import geotiff_loader
```

usl_pipeline/tests/formats/test_geotiff_loader.py
```diff
@@ -0,0 +1,35 @@
...
+                [5.648889064788818, 5.72777795791626],
+                [5.713333606719971, 5.742222309112549],
+            ],
+        )
+        self.assertEqual(crs, CRS({"init": "EPSG:32618"}))
+        self.assertFalse(crs.is_geographic)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
waltaskew, line 35: There's also no need for a main in pytest style. See my example here:
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_pipeline/tests/test_adder.py

waltaskew (changes requested):

waltaskew:
A question on modules & layout. What other functions are you planning on adding to the `geotiff_loader` module? We wouldn't want a module with just one `load_elevation_from_geotiff` function in it. Might suggest something like a single
`usl_pipeline/readers.py` and sticking all the `load_X` (or `read_X`) functions in there.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
+import rasterio.features
+
+
+def load_elevation_from_geotiff(
+    file_path: str, load_data: bool = False, result_nodata: float = None
```
waltaskew, line 5: When we use this function, I'd expect the data will be in GCS rather than on the local filesystem, so a file path won't be helpful in GCS. I'd have the function take a file / file-like object so it can function in both contexts.

rsutormin: The next step after reading header from GeoTIFF will be to cut giant TIFF into smaller chunks using GDAL utility which cannot work with remote file-like blobs on GCS. So I think we would have to load this file to local FS first. WDYT?

waltaskew: I'm saying that if the interface is a file-like object, we can handle both use cases.
If we have

```python
def load_elevation_from_geotiff(file: typing.BinaryIO, ...):
```
we can call it with
```python
with open(file_path) as fd:
  load_elevation_from_geotiff(fd)
```
or with
```python
with bucket.blob(gcs_path).open() as fd:
  load_elevation_from_geotiff(fd)
```

waltaskew: Once the chunks are created, we'll need to read the tiff chunk in the code preparing the feature matrix and this function could be useful in that context.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Yeah, I added "band" parameter with 1 as a default.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Switched to numpy NDArray.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Yeah, removed those (and switched to data object for header instead of dict).

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Done

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: I decided to follow https://google.github.io/styleguide/pyguide.html#383-functions-and-methods. Let me know if you prefer different style.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: I went with 'header_only=false' option.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Done

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
```diff
@@ -0,0 +1,39 @@
...
+):
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/tests/formats/test_geotiff_loader.py
rsutormin: Done

usl_pipeline/tests/formats/test_geotiff_loader.py
rsutormin: Done

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Done

.gitignore
```diff
@@ -1,4 +1,6 @@
+.idea
```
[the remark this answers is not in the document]
rsutormin: Done

rsutormin:
Ready for next round

usl_pipeline/usl_pipeline/readers/__init__.py
waltaskew, line 1: I think in python3 we shouldn't need `__init__.py` files any longer.
We need the one in the base `usl_pipeline` directory just so `setuptools.find_packages` in `setup.py` can find the package.

usl_pipeline/dev-requirements.txt
```diff
@@ -4,10 +4,12 @@ flake8==7.0.0
 iniconfig==2.0.0
 mccabe==0.7.0
 mypy-extensions==1.0.0
+numpy==1.26.4
```
waltaskew, line 7: We want to include the full set of dependencies -- the full transitive closure -- in the requirements file.
We want to have just top-level dependencies, like numpy and rasterio, in setup.py
I showed how to do this over here:
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/6/files

You can use `pip freeze > requirements.txt` to dump the full set of installed packages.

We want to state the full set of dependencies in a requirements file to avoid pulling in different versions of the dependencies-of-dependencies every time we pip install them.

The requirements in setup.py are meant to be looser so it's easier to install as one of many libraries which might require more certain versions of various packages.

usl_pipeline/tests/readers/test_data.tif
waltaskew, line 1: Over here:
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/7/files
I figured out how to test tiff reading functionality without relying on actual files.
I like that pattern so we can avoid filling up the test directory with various tiff files for various tests cases.

usl_pipeline/usl_pipeline/readers/elevation.py
```diff
@@ -0,0 +1,54 @@
...
+            nodata_value=final_nodata_value,
+            crs=src.crs,
+        )
+        elv_data = None
+
+        if not header_only:
+            data = src.read(band)
+            if no_data_value is not None:
+                data[data == input_nodata] = no_data_value
+            elv_data = data.tolist()
```
waltaskew, line 52: Looks like we're still returning a list rather than a numpy array here. Is that intentional?

usl_pipeline/tests/readers/test_elevation.py
```diff
@@ -0,0 +1,49 @@
...
+        assert elevation.header == ElevationHeader(
+            col_count=2,
+            row_count=2,
+            x_ll_corner=587069.5669010617,
+            y_ll_corner=4505948.133973204,
+            cell_size=2.0,
+            nodata_value=0.0,
+            crs=CRS({"init": "EPSG:32618"}),
+        )
+        assert elevation.data == [
```
waltaskew, line 24: Can use
https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html
for numpy arrays

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
waltaskew, line 1: Did you mean to remove this file? Looks like it's replaced by readers/elevation.py

usl_pipeline/usl_pipeline/shared/entities.py
waltaskew, line 1: `shared/entities` feel a little general -- 'entities' feel a bit like 'util' as a name in that it can end up being a repository for most anything.
Maybe something like
`readers/geo` ?  `geo_info`  `geo_data` ?

usl_pipeline/usl_pipeline/shared/entities.py
```diff
@@ -0,0 +1,36 @@
...
+    y_ll_corner: float
+    cell_size: float
+    nodata_value: float
+    crs: CRS = None
+
+
+@dataclass
+class Elevation:
+    header: ElevationHeader
+    data: npt.NDArray[np.float64] = None
```
waltaskew, line 36: We should add some type checking to the build to catch things like this.
The type here should be `npt.NDArray[np.float64] | None` or equivalently `Optional[npt.NDArray[np.float64]]` since it can be `None`
the type checking would also catch that we're setting this attribute to a list with the `tolist` call above.

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
waltaskew: Works for me as long as it's compatible with black and flake8 and all. 
Can you add a line or two to
https://www.notion.so/Software-Development-Life-Cycle-05b5181505cd4cd298299b6cd6c745ad
about this style guide?

usl_pipeline/usl_pipeline/readers/elevation.py
```diff
@@ -0,0 +1,54 @@
...
+
+from usl_pipeline.shared.entities import Elevation
+from usl_pipeline.shared.entities import ElevationHeader
+
+
+def read_from_geotiff(
+    file: typing.BinaryIO,
+    header_only: bool = False,
+    band: int = 1,
+    no_data_value: float = None,
```
waltaskew, line 13: ah, likewise, the type here is `float | None` or `Optional[None]`

waltaskew:
Back to you - I like the structure of the Elevation & ElevationHeader struts!
Looks like some of my comments are here on 
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/3
rather than inside
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/3/files
because I made some comments on files themselves rather than on a line within the file. I'll avoid doing that in the future!

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Done

usl_pipeline/usl_pipeline/readers/__init__.py
rsutormin: Removed them

usl_pipeline/usl_pipeline/readers/elevation.py
rsutormin: Done

usl_pipeline/tests/readers/test_elevation.py
rsutormin: Done

usl_pipeline/usl_pipeline/readers/elevation.py
rsutormin: Done

usl_pipeline/usl_pipeline/shared/entities.py
rsutormin: Done

usl_pipeline/usl_pipeline/shared/entities.py
rsutormin: geo_data works for me.

usl_pipeline/tests/readers/test_data.tif
rsutormin: Done

usl_pipeline/dev-requirements.txt
rsutormin: Done

usl_pipeline/usl_pipeline/formats/geotiff_loader.py
rsutormin: Done

waltaskew (approved):
Hooray for first commit!

PR #4, opened 2024-03-25
create the interface for our chunk upload script

waltaskew:


usl_pipeline/usl_pipeline/upload_map.py
```diff
@@ -0,0 +1,39 @@
+import argparse
+
+
+def main() -> None:
+    """Breaks the input files into chunks and uploads them to GCS."""
+    parser = argparse.ArgumentParser(
+        description=(
+            "Uploads a set of files describing features of a geography into GCS. The "
```
rsutormin, line 8: Is this script going to be used for uploading a map like NYC for future predictions based on a model that was already trained or this is going to be an entry point map for training process. I guess we need to know this already when you define arguments because when this script produces .tar files into GCS it's too late to specify what to do next (since your idea was to automatically start processing as serverless functions triggered by .tar files), right?

waltaskew: It should be the same script for both cases. Whether you want to use the map for predictions or training, we need to produce the same features matrices using the same process.

The only difference is for a map that's purely for prediction, we can optionally skip the step where we produce CityCAT inputs. Or we can produce them anyway in case folks decide they want to run CityCat on that map in the future.

waltaskew: > I think we need to have some arguments here for firebase connection.
My hope was that the firebase entries are something we'd write in the GCP trigger functions rather than from this script. That way, the user running this doesn't actually need firebase write access, just GCP write access.

> This script describes purely flooding use case, right? Should we name the script more specifically?
We can discuss if this is a good idea, but I was actually imagining it as a script for both. It's really just a difference of whether the user is setting an `--elevation-file` or an `--atmospheric-info-file` or whatever it is the atmospheric process needs. Then it's the pipeline's job to now how to turn each of those files into features. My hope is that for files that are common between the two, we'd apply identical processing rather than need to, say, process elevation one way for flood and another way for atmospheric.

rsutormin: So you don't like the idea of having top-level entity in meta-store for a map? This script is the best place where such entity should be added to the meta-store.

waltaskew: > So you don't like the idea of having top-level entity in meta-store for a map? This script is the best place where such entity should be added to the meta-store.

Oh, I do want a top level entry for map! My suggestion is just that we can create that entry in a cloud function that's triggered when we upload data to the 'map' bucket rather than something the file uploader & chunker would initiate.

If that isn't feasible we can initiate it from the uploader & chunker script. My preference would just be to avoid that for permissions reasons. I'd like to not have to grant write privileges to the Metastore to all users who want to run the upload & chunk script.

It's also nice for consistency reasons. Initiating that Metastore write from a cloud function helps reduce edge cases where the GCS write completes but the metastore write does not.

rsutormin: Yeah, ok, sorry for sitting on this concept for quite a while. Now it starts being more clear for me that this map uploader is triggered by initial placement of the tar with all the inputs for a given map. And then this uploader would prepare and place chunk tars that will trigger chunk processor functions. Finally!

rsutormin (approved):

PR #6, opened 2024-03-25
add rasterio requirements & fix tests

waltaskew:


rsutormin (commented):
I think I fixed all these problems in https://github.com/UrbanSystemsLab/climateiq-cnn/pull/3 (for instance I switched to memory file in tests as we discussed there). Thanks for sharing this example though!

waltaskew:
> I think I fixed all these problems in #3 (for instance I switched to memory file in tests as we discussed there). Thanks for sharing this example though!

cool -- I'll close it
I made this as a PR into your branch, so if you'd wanted to, you could have merged this code directly into your branch.

PR #7, opened 2024-03-27
start building out the gcp transforms

waltaskew:


usl_pipeline/usl_pipeline/transforms.py
```diff
@@ -0,0 +1,51 @@
...
+def _build_feature_matrix_from_archive(archive: BinaryIO) -> numpy.matrix | None:
+    """Builds a feature matrix for the given archive."""
+    with tarfile.TarFile(fileobj=archive) as tar:
+        for member in tar:
+            name = pathlib.PurePosixPath(member.name).name
+            if name == "elevation.tiff":
+                fd = tar.extractfile(member)
+                with rasterio.open(rasterio.io.MemoryFile(fd.read())) as raster:
+                    return raster.read(1)
+            # TODO: handle additional archive members.
```
rsutormin, line 49: Do you envision this logic to be universal for all the simulation types (flooding, heat, and so on) or this code is bound to flood only? For the first option we need some indication what simulation type (and consequently what input structure) it is from the tar content? For the second option we might need to rename (or repackage) this file.

waltaskew: So to summarize the conversation we just had about this for posterity:
- my current assumption is that we'll process the same files the same way, regardless of simulation type
- if that assumption turns out to be incorrect, we'd add a second trigger function for that simulation type and upload inputs to a different bucket for that simulation type (while trying to share code between the two triggers functions as we can)

rsutormin: Ok, sounds good.

rsutormin (approved):

PR #11, opened 2024-03-29
oh no, a bug!

waltaskew:


usl_pipeline/usl_pipeline/adder.py
```diff
@@ -1,2 +1,6 @@
 def add(x, y):
     return x + y
+
+
+def sub(x, y):
```
waltaskew, line 5: I think you could add some documentation here

usl_pipeline/usl_pipeline/adder.py
```diff
@@ -1,2 +1,6 @@
...
+    return x - y
```
waltaskew, line 6: great job with the subtraction! what a function!

waltaskew: thanks :)

usl_pipeline/usl_pipeline/adder.py
waltaskew: maybe later...

ahmustafa:
good to go

ahmustafa:
good

PR #13, opened 2024-03-29
Create File Chunk Trigger

waltaskew:
- Create a cloud function to process file chunks into feature matrices.


cloud_functions/generate_feature_matrix/main.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Write the feature matrix in .npy format to GCS.
+    matrix_file = io.BytesIO()
+    numpy.save(matrix_file, feature_matrix)
+    # Seek to the beginning so the file can be read.
+    matrix_file.seek(0)
+    feature_blob.upload_from_file(matrix_file)
+
+
+def _build_feature_matrix_from_archive(archive: BinaryIO) -> numpy.matrix | None:
```
rsutormin, line 57: Since this part doesn't depend on cloud functions environment, can we move this to our main code base living in usl_pipeline? I think it would be convenient in terms of possibility to reuse this part in local en-to-end runs based on local files/directories.

waltaskew: Yes, I'd like to do that.
It's going to take a minute to figure out how to package up the usl_pipeline library with the cloud function. I think that might make sense as a follow-up PR.

rsutormin: Sure, let's do in a separate one. No worries.

rsutormin (approved):

PR #14, opened 2024-04-02
Write Feature Chunk Creation to Firestore

waltaskew:


cloud_functions/generate_feature_matrix/main.py
```diff
@@ -66,3 +74,21 @@ def _build_feature_matrix_from_archive(archive: BinaryIO) -> numpy.matrix | None
...
+
+
+def _write_feature_metastore_entry(feature_file_name: pathlib.Path) -> None:
+    """Writes a Firestore entry for the new feature matrix chunk."""
+    db = firestore.Client()
+    # The GCS paths should be in the form map_name/chunk_name.npy
+    map_name = str(feature_file_name.parent)
+    chunk_name = feature_file_name.stem
+    chunk_ref = (
+        db.collection("maps")
```
rsutormin, line 86: I think we need to have entity class in USL for any entity you want to read/write from/to some firestore collection. Let's work on a bridge between USL and cloud_functions. Or is this PR purely informational?

waltaskew: I'm down to put together some classes for this, although how we use them will probably be a bit different than typical in ORM land. I think each part of the pipeline will be writing an attribute or two to the document rather than an entire document at a time.

waltaskew:
obviated by https://github.com/UrbanSystemsLab/climateiq-cnn/pull/19

PR #16, opened 2024-04-03
First version of GeoTIFF elevation chunker

rsutormin:


.github/workflows/python-app.yml
```diff
@@ -24,6 +24,10 @@ jobs:
         python-version: "3.11"
     - name: Install dependencies
       run: |
+        sudo apt-get update -qq
+        sudo add-apt-repository ppa:ubuntugis/ppa -y
```
rsutormin, line 28: Adding "sudo add-apt-repository ppa:ubuntugis/ppa -y" helps a bit but not completely. It bumps up the GDAL version from 3.4.1 to 3.6.4 (which is from about a year ago)!

usl_pipeline/usl_pipeline/chunkers/elevation_chunkers.py
```diff
@@ -0,0 +1,65 @@
+from osgeo import gdal
+
+from usl_pipeline.readers import elevation_readers
+from usl_pipeline.shared import suppliers
+
+
+class GeoTiffChunker:
```
waltaskew, line 7: Do you think we need a class here, or could we get away with a top-level function that's 
```python
def split_into_chunks(
    chunk_file_path_generator: suppliers.ChunkFilePathGenerator,
    elevation_file_path: str,
    chunk_size: int)
```

usl_pipeline/usl_pipeline/chunkers/elevation_chunkers.py
```diff
@@ -0,0 +1,65 @@
...
+        col_start = x_chunk_index * self.chunk_size
+        col_count = min(self.global_col_count - col_start, self.chunk_size)
+        ds = gdal.Open(self.elevation_file_path)
+        gdal.Translate(
+            chunk_file_path, ds, srcWin=[col_start, row_start, col_count, row_count]
+        )
+
+    def split_into_chunks(
+        self,
+        chunk_file_path_generator: suppliers.ChunkFilePathGenerator,
```
waltaskew, line 52: This interface feels like it's exposing some internal concerns to callers.
I'd hope we could get
```python
def split_info_chunks(file: typing.BinaryIO):
```
or 
```python
def split_info_chunks(file_path: str | pathlib.Path):
```
rather than ask callers to need to understand how to work with `suppliers.ChunkFilePathGenerator`. 

`ChunkFilePathGenerator` feels more like an internal interface than a public one to me

usl_pipeline/tests/chunkers/test_elevation_chunkers.py
```diff
@@ -0,0 +1,165 @@
...
+        width=3,
+        height=3,
+        count=1,
+        crs=rasterio.CRS.from_epsg(32618),
+        transform=rasterio.Affine(2.0, 0.0, 10.0, 0.0, -2.0, 20.0),
+    ) as raster:
+        raster.write(tiff_array.astype(rasterio.float32), 1)
+
+
+class MemoryChunkPathGenerator(suppliers.ChunkFilePathGenerator):
```
waltaskew, line 37: What is the `MemoryChunkPathGenerator` callers would use outside of tests?
I'm concerned the tests will be testing the implementation of `MemoryChunkPathGenerator` rather than the implementation callers will actually be using.

I'd expect the non-test implementation would either expect a `file: typing.BinaryIO`  or a `file_path: str | pathlib.Path`. I'd rather work with that implementation itself in the tests to make sure we're testing the implementation real callers will actually use, even if that means having to work more with temporary files than memory files.

.github/workflows/python-app.yml
rsutormin: Done

usl_pipeline/tests/chunkers/test_elevation_chunkers.py
```diff
@@ -0,0 +1,165 @@
...
+        return "/vsimem/{}/chunk_{}_{}.tif".format(
+            self.memory_root, y_chunk_index, x_chunk_index
+        )
+
+
+def get_chunk_elevation_from_memory(
+    chunk_path_generator: MemoryChunkPathGenerator,
+    y_chunk_index: int,
+    x_chunk_index: int,
+):
```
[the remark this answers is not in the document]
rsutormin: Switched to prod implementation of the generator.

usl_pipeline/tests/chunkers/test_elevation_chunkers.py
```diff
@@ -0,0 +1,165 @@
...
+    file_path = chunk_path_generator.generate(y_chunk_index, x_chunk_index)
+    stat = gdal.VSIStatL(file_path, gdal.VSI_STAT_SIZE_FLAG)
+    # open memory file
+    vsi_file = gdal.VSIFOpenL(file_path, "r")
+    # read entire contents
+    byte_content = gdal.VSIFReadL(1, stat.size, vsi_file)
+    # write bytes to temporary file in order for rasterio library to read it
+    with tempfile.NamedTemporaryFile() as tmpfile:
+        with open(tmpfile.name, "wb") as output_file:
```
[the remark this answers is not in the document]
rsutormin: I removed this code since I moved to prod implementation of the generator.

usl_pipeline/tests/chunkers/test_elevation_chunkers.py
```diff
@@ -0,0 +1,165 @@
...
+            output_file.write(byte_content)
+        with open(tmpfile.name, "rb") as input_file:
+            return elevation_readers.read_from_geotiff(input_file)
+
+
+def check_chunk_data(
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/usl_pipeline/chunkers/elevation_chunkers.py
```diff
@@ -0,0 +1,65 @@
...
+            elevation = elevation_readers.read_from_geotiff(
+                input_file, header_only=True
+            )
+
+        self.global_col_count = elevation.header.col_count
+        self.global_row_count = elevation.header.row_count
+        self.x_chunk_count = int((self.global_col_count + chunk_size - 1) / chunk_size)
+        self.y_chunk_count = int((self.global_row_count + chunk_size - 1) / chunk_size)
+
+    def extract_chunk(
```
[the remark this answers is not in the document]
rsutormin: I removed this method during refactoring.

usl_pipeline/usl_pipeline/chunkers/elevation_chunkers.py
```diff
@@ -0,0 +1,65 @@
...
+        col_count = min(self.global_col_count - col_start, self.chunk_size)
+        ds = gdal.Open(self.elevation_file_path)
+        gdal.Translate(
+            chunk_file_path, ds, srcWin=[col_start, row_start, col_count, row_count]
+        )
+
+    def split_into_chunks(
+        self,
+        chunk_file_path_generator: suppliers.ChunkFilePathGenerator,
+    ) -> (int, int):
```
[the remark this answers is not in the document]
rsutormin: I actually DO return number of chunks along both Y- and X- axis now.

usl_pipeline/usl_pipeline/chunkers/elevation_chunkers.py
rsutormin: Done.

usl_pipeline/usl_pipeline/chunkers/elevation_chunkers.py
rsutormin: Now there is prod (default) implementation for this generator so callers now can skip understanding part.

usl_pipeline/README.md
```diff
@@ -2,6 +2,8 @@ Getting Started
 ===============
 
 To install the package and its requirements for local development:
+- System dependencies:
+  - GDAL version 3.6.4 or later (apt-get install gdal-bin libgdal-dev)
```
waltaskew, line 6: Should folks be using the ubuntugis/ppa you're using in the CI process?

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
```diff
@@ -0,0 +1,49 @@
+import abc
+import os
+
+
+class ChunkFilePathGenerator(abc.ABC):
```
waltaskew, line 5: Do we anticipate additional implementations of this interface? If not, do we need to define the interface?

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
```diff
@@ -0,0 +1,49 @@
+import abc
```
waltaskew, line 1: Might just stick this code alongside the chunking code rather than create a new module for it as I wouldn't expect it to be used in any other contexts. This code and the chunking code feel pretty entwined.

What else do you anticipate adding to this new 'suppliers' module? Would it expand past its current contents?

Is there a more clear name that 'suppliers' you could imagine? It looks like this is more like a `path` or `path_utils` or such?

What does the 'shared' part of `usl_lib/shared/suppliers` indicate? I think of everything in 'usl_lib' as being shared -- 'lib' already implies 'shared' to me.

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
```diff
@@ -0,0 +1,49 @@
...
+
+class PatternBasedChunkFilePathGenerator(ChunkFilePathGenerator):
+    """An implementation of a chunk file path generator based on file name pattern."""
+
+    def __init__(self, dir_path: str, file_name_pattern: str):
+        """Constructor.
+
+        Args:
+            dir_path: A directory where every chunk file path will point.
+            file_name_pattern: A pattern for chunk file name that will be used in
```
waltaskew, line 30: Does this need to be variable? When would we want to vary the pattern?

usl_pipeline/usl_lib/tests/chunkers/test_elevation_chunkers.py
```diff
@@ -0,0 +1,135 @@
...
+from numpy import testing
+import rasterio
+
+from usl_lib.readers import elevation_readers
+from usl_lib.chunkers import elevation_chunkers
+from usl_lib.shared import geo_data
+from usl_lib.shared import suppliers
+
+
+def prepare_test_geotiff_elevation_memory_file(file_path: str) -> None:
```
waltaskew, line 13: It looks like this is no longer a memory file?

usl_pipeline/usl_lib/usl_lib/chunkers/elevation_chunkers.py
```diff
@@ -0,0 +1,46 @@
+from osgeo import gdal
+import typing
+
+from usl_lib.readers import elevation_readers
+from usl_lib.shared import suppliers
+
+
+def split_geotiff_into_chunks(
+    elevation_file_path: str,
```
waltaskew, line 9: expecting a `typing.BinaryIO` rather than a file path, as we do in `read_from_geotiff`, can be more flexible. For instance, we could write the tests using BytesIO objects rather than tempfiles.

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
```diff
@@ -0,0 +1,49 @@
...
+        Args:
+            y_chunk_index: Index along the Y-axis for a chunk position.
+            x_chunk_index: Index along the X-axis for a chunk position.
+
+        Returns:
+            A path to a chunk file.
+        """
+        return os.path.join(
+            self.dir_path, self.file_name_pattern.format(y_chunk_index, x_chunk_index)
+        )
```
waltaskew, line 49: I'm still not sure if we need to introduce a `ChunkFilePathGenerator` abstraction. Do we actually want or need to support multiple `ChunkFilePathGenerator`s or use different patterns for naming chunks? Would this abstraction be used in by other code paths outside the chunking in here? 

If we only need one single method of naming the chunks, I think we'd be fine with

```python
def split_geotiff_into_chunks(
    elevation_file_path: str,
    chunk_size: int,
    output_dir: str,
 ) -> List[pathlib.Path]:

def chunk_name(dir_path: str, y_chunk_index: int, x_chunk_index: int) -> str:
    return os.path.join(
            dir_path, f'chunk_{y_chunk_index}_{x_chunk_index}'
    )
```
and the other chunkers can all just call `chunk_name`

usl_pipeline/usl_lib/usl_lib/chunkers/elevation_chunkers.py
```diff
@@ -0,0 +1,46 @@
...
+) -> typing.Tuple[int, int]:
+    """Produces a grid of chunk GeoTIFF files based on input GeoTIFF file.
+
+    Args:
+        elevation_file_path: GeoTIFF file with elevation data to be split into chunks.
+        chunk_size: Size of a chunk in cells (along both X and Y axes).
+        chunk_file_path_generator: Caller provided source to generate chunk file names.
+
+    Returns:
+        Tuple (number of chunks along the Y-axis, number of chunks along the X-axis).
```
waltaskew, line 21: I think it'd be even more helpful to return the file paths for the chunks. We'd like to be able to write something like
```python
for chunk in split_into_chunks():
  upload_to_gcp(chunk)
```
so the caller doesn't have to recreate the path names.

waltaskew (approved):

usl_pipeline/README.md
rsutormin: I guess it might depend on the OS. I don't know what it needs to be in case of windows. Actually https://gdal.org/download.html gives a lot of ways to download binaries for all the recent versions. I think it's also possible (in theory) to compile GDAL from sources.

usl_pipeline/usl_lib/usl_lib/chunkers/elevation_chunkers.py
rsutormin: gdal.Open() can only work with string path: https://gdal.org/api/python/raster_api.html#osgeo.gdal.Open

rsutormin: If I try to pass file reader I see the following error:
```
usl_pipeline/usl_lib/usl_lib/chunkers/elevation_chunkers.py:32: in split_geotiff_into_chunks
    ds = gdal.Open(file)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<_io.BufferedReader name='/tmp/tmp6p9b5_7d'>,)

    def Open(*args) -> "GDALDatasetShadow *":
        r"""Open(char const * utf8_path, GDALAccess eAccess=GA_ReadOnly) -> Dataset"""
>       return _gdal.Open(*args)
E       RuntimeError: not a string
```

waltaskew: cool -- let's stick with file path strings, then.

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
rsutormin: We have pretty different understanding of comfortable amount of flexibility that different parts of the logic should offer to each other. My perspective is that we might not know what needs would appear in the future for each particular sub-system. So what I'm trying to achieve is some level of independence with some way of freedom to get results in more flexible form to avoid some conflicts. Example: all our chunkers would want to produce chunk files with the same "chunk_{y}_{x}" names which would require from caller code to make different temp folders for each type of data (elevation data couldn't be processed in the same folder as say buildings data). But that's fine I can follow your suggestion now and we can reconsider it later if we need.

usl_pipeline/usl_lib/usl_lib/chunkers/elevation_chunkers.py
rsutormin: In order to avoid having the same names for different types of inputs (elevation, buildings and so on) I added optional file pattern parameter and also each chunk is now represented by not just string or path but more like a descriptor with y- and x- index in it.

usl_pipeline/usl_lib/tests/chunkers/test_elevation_chunkers.py
rsutormin: Sorry, forgot to update it for some reason...

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
rsutormin: Removed the whole generator thingy as you suggested.

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
rsutormin: It's supposed to be shared between different chunkers. But it's not relevant anymore.

usl_pipeline/usl_lib/usl_lib/shared/suppliers.py
rsutormin: In theory the caller might want to organize chunks in some different way (for instance to place all the layers corresponding to one chunk position to its own folder). But I see that this whole amount of flexibility isn't comfortable for you. Removed it all together.

usl_pipeline/README.md
rsutormin: I decided to point to https://gdal.org/download.html link in the readme instead of "apt-get" command.

PR #19, opened 2024-04-08
Write MetaStore Entries for Study Areas and Chunks

waltaskew:
- Write a metastore entry for the study area from the map_uploader script.
- Write a metastore entry for each chunk in the build_feature_matrix cloud function.
- Add classes defining the metastore schema.


usl_pipeline/usl_lib/usl_lib/storage/metastore.py
```diff
@@ -0,0 +1,80 @@
...
+
+    name: str
+    col_count: int
+    row_count: int
+    x_ll_corner: float
+    y_ll_corner: float
+    cell_size: float
+    crs: str
+
+    def create(self, db: firestore.Client) -> None:
```
rsutormin, line 34: I think metastore should also let caller load (look-up) StudyArea object by name.

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
```diff
@@ -0,0 +1,80 @@
...
+            training and prediction.
+        error: Any errors encountered while processing the chunk.
+    """
+
+    id_: str
+    archive_path: Optional[str] = None
+    feature_matrix_path: Optional[str] = None
+    error: Optional[str] = None
+
+    def merge(self, db: firestore.Client, study_area_name: str) -> None:
```
rsutormin, line 60: The same comment as for study area: I think metastore should also let caller load (look-up) StudyAreaChunk object by parent name + chunk index position ([y,x] pair).

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
```diff
@@ -0,0 +1,80 @@
...
+        id_: An ID unique within the study are for the chunk.
+        archive_path: GCS location of an archive containing files describing the
+            geography (e.g. tiff & shape files.)
+        feature_matrix_path: GCS location of the derived feature matrix used for model
+            training and prediction.
+        error: Any errors encountered while processing the chunk.
+    """
+
+    id_: str
+    archive_path: Optional[str] = None
```
rsutormin, line 56: Can we add "y_index" and "x_index" position pair of the chunk?

rsutormin (changes requested):

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
waltaskew: It will!
I was only doing the writing in this PR.

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
waltaskew: Definitely! I'm just only looking to do writes in this PR.

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
waltaskew: To support that we could either
- use concatenated (y, x) indices as the id for the chunk
- add separate 'y_index' and 'x_index' fields and create a compound index over the two fields to query

I think the indices make for a reasonable document ID (it only needs to be unique with the parent study area, not globally over all chunks) but having explicit fields + a compound index feels a bit more legible.

But either way, I'd like to do the actual reading in another PR.

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
rsutormin: I see. SG!

rsutormin (approved):

PR #21, opened 2024-04-10
Add data utils for Flood CNN

tatianawu:


flood_ml/data_utils.py
```diff
@@ -0,0 +1,64 @@
+"""Data functions for Flood CNN model."""
+
+from constants import GEO_FEATURES, MAP_HEIGHT, MAP_WIDTH, N_FLOOD_MAPS
```
waltaskew, line 3: Google style guide sez

"Use import statements for packages and modules only, not for individual types, classes, or functions."

https://google.github.io/styleguide/pyguide.html#22-imports

which mirrors

```
>>> import this
The Zen of Python, by Tim Peters

...
Namespaces are one honking great idea -- let's do more of those!
```

flood_ml/data_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    B = data.shape[0]
+    padding = np.zeros([B, window - 1])
+    data = np.concatenate([padding, data], axis=-1)
+    return sliding_window_view(data, window, axis=-1)
+
+
+def fake_input_batch(
+    batch_size,
+    height=None,
```
waltaskew, line 30: looks like a
```python
height=MAP_HEIGHT
```
here would cover what you're doing below with
```python
height = height or MAP_HEIGHT
```

flood_ml/data_utils.py
```diff
@@ -0,0 +1,64 @@
...
+        A 3D tensor of shape [B, T, window]. In order to maintain the same T,
+        we use zero padding at the beginning of the temporal data.
+    """
+    B = data.shape[0]
+    padding = np.zeros([B, window - 1])
+    data = np.concatenate([padding, data], axis=-1)
+    return sliding_window_view(data, window, axis=-1)
+
+
+def fake_input_batch(
```
waltaskew, line 28: Is this function useful outside of tests? If not, I'd stick it somewhere like
`flood_ml/testing/data_utils.py`
or
`tests/data_utils.py`
and perhaps giving it a more specific name than 'utils' like `data_gen.py`

flood_ml/data_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    batch_size,
+    height=None,
+    width=None,
+    rainfall_duration=24,
+    flood_maps=None,
+    include_flood_maps=False,
+):
```
waltaskew, line 35: Let's add a type annotations for the return type and some information about what we return to the docstring.

flood_ml/data_utils_test.py
```diff
@@ -0,0 +1,37 @@
+import data_utils
+
+from numpy.testing import assert_array_equal
+import tensorflow as tf
+import unittest
+
+
+class TestDataUtils(unittest.TestCase):
```
waltaskew, line 8: We've been writing pytest style tests where a test class and a unittest.main() call are unnescessary.

https://docs.pytest.org/en/8.0.x/#a-quick-example
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_pipeline/cloud_functions/main_test.py

flood_ml/data_utils_test.py
```diff
@@ -0,0 +1,37 @@
+import data_utils
```
waltaskew, line 1: We'll need to update the GitHub CI to actually run these tests (and linters and so on)
It should look a lot like:
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/.github/workflows/python-app.yml#L44

flood_ml/constants.py
```diff
@@ -0,0 +1,10 @@
+"""Constant definitions for Flood CNN."""
```
waltaskew, line 1: Some notes on package structure. I'd suggest a directory structure like
```
flood_ml/
  - requirements.txt
  - setup.py
  - flood_ml/
    - __init__.py
    - constants.py
    - data_utils.py
  - tests/
    - test_data_utils.py
```
That will be compatible with the conventions tools like pytest use for finding tests and create a clearly defined `flood_ml` package with attendant requirements.txt and/or setup.py files.

flood_ml/data_utils.py
tatianawu: Fixed.

flood_ml/data_utils.py
tatianawu: Good call. Made the change, but I removed this function from the PR.

flood_ml/data_utils.py
tatianawu: I wrote this function with testing in mind, but it's currently not being used, so I removed it from this PR.

flood_ml/data_utils.py
tatianawu: Sorry, I missed the return statement in the docstring. Do you want type annotations for the code in general, or was this because the docstring was also missing the information?

flood_ml/data_utils_test.py
tatianawu: Fixed.

flood_ml/data_utils.py
waltaskew: I'm into type annotations generally. I do think they're useful for documentation & readability. 

We've set up MyPy to do type checking for the other packages, but it needs type annotations to do anything useful. MyPy just ignores functions when they don't have type annotations

```python
def func():
  return None

1 + func()
```
MyPy Type checking will be perfectly happy

```python
def func() -> None:
  return None

1 + func()
```
now MyPy will actually diagnose a type error

models/setup.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+
+from setuptools import setup, find_packages
+
+setup(
+    name="flood_ml",
```
waltaskew, line 6: Do you intend to have one package for each model? Or one package containing both models? Right now, we still have one package named flood_model. You'd either want
```
models/
  - setup.py
  - models/
    -  __init__.py
    - flood_model/
    - wrf_mode/
```
for one package containing both models or
```
flood_model/
  - setup.py
  - flood_model/
    -  __init__.py
wrf_model/
  - setup.py
  - wrf_model/
    -  __init__.py 
```
for one package for each model.

I don't feel stro

flood_ml/data_utils.py
tatianawu: Sure, I'm generally a fan of annotations too :) I think with ML code it can become a bit unwieldy. I'm not sure if specifying a generic type like `tf.Tensor` is actually that useful. We could define more detailed annotations, for example using [jaxtyping](https://github.com/patrick-kidger/jaxtyping), to include information about expected shapes, which I'm currently including in docstrings instead. My main question here is whether this would be useful to TNS and whether they would want to maintain the convention post-fellowship, since these are non-standard annotations and IMO not the most intuitive.

Maybe for more general code like this function, we can include type annotations. And within the model itself, I'll see if it still makes sense.

waltaskew: Yep, I think that makes sense. Happy to leave them off where they aren't helpful.

models/setup.py
tatianawu: Sorry, what's the difference? I'm not really familiar with this packaging structure, so I don't understand how this might impact workflows down the line.

waltaskew: The first 'one package containing both models' option would result in imports like
`from models.flood_model import ...`
The second 'one package for each model' option would be
`from flood_model import ...`

The folder sitting beside the `setup.py` will become the namespace for your package (either 'model' or 'flood_model' in the imports above.)
On that note, would probably want a namespace more distinct that 'model.' 'usl_model' would match what we're doing with 'usl_lib' and 'usl_pipeline' elsewhere in here.

waltaskew: Down the line, it will determine things like whether we need to maintain two docker images when we pull things into vertex ai. I'd just recommend one package containing both models ala

```
models/
  - setup.py
  - requirements.txt
  - usl_models/
    -  __init__.py
    - flood_model/
    - wrf_mode/
```

for imports like
`from usl_models.flood_model import ...`

but we only have to maintain one requirements.txt file and one docker image for the two models.

tatianawu: Yeah, I had figured that a single package would allow us to bundle some common things like the requirements file. Is it not possible to have a structure like this?

```
usl_models/
  - setup.py
  - requirements.txt
  - flood_model/
    - __init__.py
  - wrf_model/
    - __init__.py
  - tests/
```

waltaskew: That creates two namespaces, `flood_model` and `wrf_model`. It's unusual for one package to own two different name spaces. I think what you have would work, it would just be unusual. The more common thing would be to stick flood_model & wrf_model under a single namespace and export that single namespace in your package.

flood_ml/constants.py
tatianawu: Resolving this, since there's another thread about the package structure.

flood_ml/data_utils_test.py
tatianawu: Added workflows, and looks like they passed. Can you help confirm that these are set up properly?

models/setup.py
tatianawu: Updated the structure so it's a single `usl_models` package, based on your suggestion.

.github/workflows/python-app.yml
```diff
@@ -151,3 +151,47 @@ jobs:
...
+        python-version: "3.11"
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        cd usl_models
+        pip install -r requirements.txt
+        pip install -e .[dev]
+    - name: MyPy
+      run: |
+        mypy usl_models
```
waltaskew, line 197: I'd just add the `mypy usl_models` step to the `USL Models CI` step above.
The reason I have it as separate above is to type check all the different usl_pipeline packages together. MyPy isn't as smart if it doesn't see the actual source files for imported code.

usl_models/setup.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+
+from setuptools import setup, find_packages
+
+setup(
+    name="usl_models",
+    packages=find_packages(),
```
waltaskew, line 7: might add an
```python
install_requires=["tensorflow"],
```
in here, alongside any other high-level dependencies you care about.

waltaskew (approved):
Hooray! The joys of Python packaging!

.github/workflows/python-app.yml
tatianawu: Done.

usl_models/setup.py
tatianawu: Done.

PR #22, opened 2024-04-11
Polygon reader from shape files

rsutormin:


usl_pipeline/cloud_functions/requirements.txt
```diff
@@ -9,6 +9,7 @@ click-plugins==1.1.1
 cligj==0.7.2
 cloudevents==1.10.1
 deprecation==2.1.0
+fiona==1.9.6
```
waltaskew, line 12: this will break the cloud function if it requires GDAL to be present during pip install time :'(

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
```diff
@@ -0,0 +1,67 @@
...
+
+SHAPE_FILE_POLYGON_FEATURE_TYPE = "Polygon"
+
+
+def read_polygons_from_shape_file(
+    file_path: str,
+    target_crs: typing.Optional[str] = None,
+    mask_value_feature_property: typing.Optional[str] = None,
+    skip_zero_mask_values: bool = True,
+) -> list[typing.Tuple[Polygon, int]]:
```
waltaskew, line 17: Things from typing are one of the few exceptions to the 'only import namespaces' rule:
https://google.github.io/styleguide/pyguide.html#2241-exemptions

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
```diff
@@ -0,0 +1,67 @@
...
+    """Reads polygon data from shape file.
+
+    Function reads all the polygon shapes from shape file. There is optional support for
+    transformation of coordinates to different CRS. Another option is to load masks
```
waltaskew, line 21: When you write 'Another option is' that read more like, 'An alternative option is,' as in, an alternative to the first option rather than a second, additional optional parameter. I'd write this as, 

"The function supports several optional transformations. `target_crs` transforms the coordinates to a different CRS. `mask_value_feature_property` loads masks ...

Or alternatively, just rely on the below 'Args:' section to describe these.

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
```diff
@@ -0,0 +1,67 @@
...
+    for feature in layer:
+        feature_type = feature.geometry.type
+        if feature_type != SHAPE_FILE_POLYGON_FEATURE_TYPE:
+            continue
+        fragments = feature.geometry.coordinates
+        mask_value = (
+            1
+            if mask_value_feature_property is None
+            else int(
+                fiona.model.to_dict(feature.properties)[mask_value_feature_property]
```
waltaskew, line 54: might consider re-raising the exception for a bad property as something clearer
```python
try:
  property = fiona.model.to_dict(feature.properties)[mask_value_feature_property]
except KeyError:
  raise ValueError(
    f'Invalid mask_value_feature_property {mask_value_feature_property},  '
    f'must be one of {", ".join(fiona.model.to_dict())}`

File: usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py [hunk 5]
```diff
@@ -0,0 +1,67 @@
...
+            )
+        )
+        if mask_value == 0 and skip_zero_mask_values:
```
waltaskew, line 57: Does this only trigger if the caller's `mask_value_feature_property` turns out to be mask_value 0?
Why would we want to skip it -- what does a mask value of 0 represent? I think the reason behind skipping 0 mask values would be helpful to add to the docstring.

waltaskew (commented):
The code all LGTM, just some documentation suggestions, but I'm worried this will break the cloud function if it starts depending upon Fiona.

usl_pipeline/cloud_functions/requirements.txt
waltaskew: okay! We talked and @rsutormin , you correctly pointed out that it would break the build if this was an issue!

We might still want to move the shape files readers out of `usl_lib` and into `map_uploader` because the cloud function would break if it ever attempted to call the shape file reader, as GDAL won't be present.

waltaskew (approved):

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
rsutormin: Done

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
rsutormin: Done

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
rsutormin: Done. I also added the check for parsing int value.

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
rsutormin: We use 0 values during rasterization as a background after these masks are loaded. I noticed that some polygons in soil classes shape-file for NYC have 0 masks and overlap with other polygons leading to "drawing" background values over other soil masks (basically clearing out the soil info for many raster cells). Actually I think I can move this filtering logic out of reading function in order to make this low-level read operation clearer.

usl_pipeline/cloud_functions/requirements.txt
rsutormin: I moved this code to map_uploader and removed dependency to fiona lib from usl_lib and from cloud_functions.

PR #23, opened 2024-04-12
Implement forward pass of Flood ConvLSTM model

tatianawu:
* Defined the ConvLSTM architecture and implemented the logic for a single forward pass of the model.
* This model is not trainable yet since `call` is unimplemented. This will be added with the autoregression logic in a future step.


tatianawu:
Sorry about the weird commits. I think this is because I did a squash on the last PR and then merged main into the same branch.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -0,0 +1,129 @@
...
+        # Spatiotemporal CNN
+        st_cnn_params = {"strides": 2, "padding": "same", "activation": "relu"}
+        st_cnn_layers = [
+            layers.Conv2D(16, 5, **st_cnn_params),
+            layers.MaxPool2D(pool_size=2, strides=1, padding="same"),
+            layers.Conv2D(64, 5, **st_cnn_params),
+            layers.MaxPool2D(pool_size=2, strides=1, padding="same"),
+        ]
+        self.st_cnn = tf.keras.Sequential(
+            [layers.TimeDistributed(layer) for layer in st_cnn_layers]
```
waltaskew, line 45: It looks like Sequential will accept a generator, so you can avoid materializing that list by removing those brackets
```python
 self.st_cnn = tf.keras.Sequential(
    layers.TimeDistributed(layer) for layer in st_cnn_layers
)

**Review by waltaskew — APPROVED** [review]
Love all the documentation!

File: usl_models/tests/model_test.py [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+    temp_input = data_utils.temporal_window_view(
+        fake_input["temporal"], _MODEL_PARAMS["m_rainfall"]
+    )
+    # The forward function assumes temp_input has been clipped to the same
+    # time slice as the flood maps. We take the first <flood_maps> values.
+    temp_input = temp_input[:, :flood_maps, :]
+
+    model = flood_model.FloodConvLSTM(_MODEL_PARAMS)
+    prediction = model.forward(st_input, geo_input, temp_input)
+    assert prediction.shape == (batch_size, height, width, 1)
```
SallyElHajjar, line 84: I like the structure and the explanation of each block

SallyElHajjar (approved):

flood_ml/requirements.txt
```diff
@@ -0,0 +1,41 @@
...
+certifi==2024.2.2
+charset-normalizer==3.3.2
+click==8.1.7
+flatbuffers==24.3.25
+gast==0.5.4
+google-pasta==0.2.0
+grpcio==1.62.1
+h5py==3.10.0
+idna==3.6
+keras==3.1.1
```
SallyElHajjar: Even if Tensorflow versions come with built-in Keras, should we still specify keras as requirement?

usl_models/usl_models/flood_ml/model.py
tatianawu: Done.

flood_ml/requirements.txt
tatianawu: I didn't install Keras separately. The TF install includes Keras and many other packages, and `pip freeze` includes all of them.

PR #24, opened 2024-04-15
Remove usl_pipeline/requirements.txt

waltaskew:
- This isn't a real requirements.txt tied to a real binary, so let's remove it rather than confuse folks.


rsutormin (approved):

.github/workflows/python-app.yml
```diff
@@ -44,6 +40,9 @@ jobs:
     - name: Test with pytest
       run: |
         pytest usl_pipeline/usl_lib
+    - name: MyPy Type Checking
+      run: |
+        pytest usl_pipeline/usl_lib
```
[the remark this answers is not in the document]
waltaskew: It should!

.github/workflows/python-app.yml
```diff
@@ -72,7 +71,10 @@ jobs:
         black usl_pipeline/map_uploader --check
     - name: Test with pytest
       run: |
-        pytest usl_pipeline/map_uploader
+        pytest usl_pipeline/map_uploader usl_pipeline/usl_lib
+    - name: MyPy Type Checking
+      run: |
+        mypy usl_pipeline/map_uploader usl_pipeline/usl_lib
```
rsutormin, line 77: Can you do separate checks like
```
mypy usl_pipeline/map_uploader
mypy usl_pipeline/usl_lib
```
(otherwise I have mypy error like in https://github.com/UrbanSystemsLab/climateiq-cnn/actions/runs/8695726630/job/23847381930?pr=25)?

rsutormin (changes requested):

.github/workflows/python-app.yml
waltaskew: Done. We don't really need to MyPy both anyway here.

rsutormin (approved):

PR #25, opened 2024-04-15
Adding polygons reader from text file

rsutormin:


usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
```diff
@@ -0,0 +1,49 @@
...
+    Args:
+        file: File text stream to load from.
+        support_mask_values: Optional indicator of the presence of polygon mask values
+            in the first column in polygon lines. Default case is that mask column is
+            not present (in this case mask value 1 will be assigned to every polygon).
+
+    Returns:
+        The list of tuples combining polygon with associated mask.
+    """
+    lines = file.readlines()
```
waltaskew, line 21: That's going to read the entire file into memory at once.
Can we process this file line by line instead?

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def read_polygons_from_text_file(
+    file: typing.TextIO, support_mask_values: bool = False
+) -> list[Tuple[geometry.Polygon, int]]:
+    """Reads polygon information with optional mask values from text file.
+
+    Args:
+        file: File text stream to load from.
+        support_mask_values: Optional indicator of the presence of polygon mask values
```
waltaskew, line 14: Should we handle this automatically? As in, when reading the file, check to see if mask values are present, and if so returning them? Is there a reason why callers would want to have support_mask_values as False, even if mask values are present?

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
```diff
@@ -0,0 +1,49 @@
...
+            index_base = 1
+        pnt_count = int(parts[index_base])
+        points = []
+
+        for j in range(0, pnt_count):
+            x = float(parts[index_base + 1 + j])
+            y = float(parts[index_base + 1 + pnt_count + j])
+            points.append((x, y))
+
+        polygons.append((geometry.Polygon(points), mask_value))
```
waltaskew, line 47: Might consider making this a generator instead of materializing the list of all polygons. Do we need to materialize that list, or will callers be working polygon by polygon, and thus only need an iterator rather than a list?

We'd just do
```python
yield (geometry.Polygon(points), mask_value)
```
here
and then callers would use this function like
```python
for polygon in read_polygons_from_text_file(...)
```
although it'd be more common to name the function `iter_polygons_from_text_file` or such.

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
rsutormin: There is no need because shape files are relatively small and you'll get all the polygon objects in memory anyway.

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
rsutormin: Sure. Let me do it...

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
rsutormin: I don't see any benefits in iteration along the generator over iteration along the normal list. The data in shape files is relatively small comparing to Elevation. Splitting logic that will come a few PRs later keeps all the polygons in memory anyway. Though if we're not satisfied with the memory usage we might reconsider this approach later. But I don't see any reason to spend time on optimization right now.

waltaskew (approved):

usl_pipeline/usl_lib/usl_lib/readers/polygon_readers.py
```diff
@@ -0,0 +1,47 @@
+import typing
+from typing import Generator, Tuple
+
+from shapely import geometry
+
+
+def read_polygons_from_text_file(
+    file: typing.TextIO,
+) -> Generator[Tuple[geometry.Polygon, int], None, None]:
```
waltaskew, line 9: We can use a simpler / more general type here like `Iterable[Tuple[geometry.Polygon, int]]`
since we aren't doing fancy stuff with generators like sending values back into the generator.

 I think the iterable type is a little more readable, but either way is fine.

from the docs:
https://docs.python.org/3.6/library/typing.html#typing.Generator
"Alternatively, annotate your generator as having a return type of either Iterable[YieldType] or Iterator[YieldType]:"

waltaskew (approved):

PR #26, opened 2024-04-17
Polygon writer to text format

rsutormin:


usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
```diff
@@ -0,0 +1,40 @@
+import typing
+from typing import Tuple
+
+from shapely import geometry
+
+
+def num_list_to_strings(num_list):
+    return list(map(lambda v: str(v), num_list))
```
waltaskew, line 8: the `" ".join` calls below will happily take the generator returned by `map`, so we can drop the `list` call here to avoid materializing a list.
Can also drop the lambda so this could just be
```python
return map(str, num_list)
```
Python folks also like to write this as comprehensions, so below we could also do
```python
" ".join(str(v) for v in num_list)
```

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
```diff
@@ -0,0 +1,40 @@
...
+            as additional first column.
+
+    Returns:
+        The generator of tuples combining polygon with associated mask to iterate over.
+    """
+    output_file.write(f"{len(polygons_with_masks)}\n")
+    for polygon_mask in polygons_with_masks:
+        xx, yy = polygon_mask[0].exterior.coords.xy
+        x_list = xx.tolist()
+        y_list = yy.tolist()
```
waltaskew, line 31: Why do we need the `tolist` calls here rather than iterating over `polygon_mask[0].exterior.coords.xy.xx` & `xy` directly?

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
```diff
@@ -0,0 +1,40 @@
...
+        mask_prefix = f"{polygon_mask[1]} " if support_mask_values else ""
```
waltaskew, line 32: Similar to the previous PR, can we detect if a mask is present here versus asking the caller to pass  `support_mask_values`?

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
```diff
@@ -0,0 +1,40 @@
...
+
+from shapely import geometry
+
+
+def num_list_to_strings(num_list):
+    return list(map(lambda v: str(v), num_list))
+
+
+def write_polygons_to_text_file(
+    polygons_with_masks: list[Tuple[geometry.Polygon, int]],
```
waltaskew, line 12: I think this is an `Iterable` now rather than a `list`?

waltaskew (approved):
Just some minor things, hooray for polygons!

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
rsutormin: This time it's not about detection and more like about preparing textual file with polygons that CityCat program can use. I want to be able to define precisely whether we want to have a column with masks in the polygon file or not (for instance for buildings we shouldn't have one and for green areas we may or may not depending on the soil info presence).

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
```diff
@@ -0,0 +1,40 @@
...
+    Args:
+        polygons_with_masks: List of tuples combining polygon with associated mask.
+        output_file: File text stream to write to.
+        support_mask_values: Optional indicator that masks should be added into output
+            as additional first column.
+
+    Returns:
+        The generator of tuples combining polygon with associated mask to iterate over.
+    """
+    output_file.write(f"{len(polygons_with_masks)}\n")
```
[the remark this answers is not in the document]
rsutormin: Sorry, no, this function isn't supposed to return anything. I copy-pasted this part and forgot to remove the "Returns" line. Removed it.

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
rsutormin: Oh, you're right, I don't actually need to switch to list and to have the "num_list_to_strings" function at all! Removed it completely.

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
rsutormin: Yep, that's right, I removed the "tolist" and the "num_list_to_strings" all together.

usl_pipeline/usl_lib/usl_lib/writers/polygon_writers.py
rsutormin: It's not convenient for me here to deal with iterable here since I need to write the number of polygons in the first output line. Otherwise I would have to convert iterable to the list internally which I don't want to do (it's better to do it outside of the function and might not be needed at all for cases when the caller already has a list to send as an input). As I mentioned before, the splitting logic in chunker collects lists of polygons for each chunk and that's fine (not too much memory is needed for keeping all the polygons in memory at the same time). So let's not over-complicate the logic here?

PR #31, opened 2024-04-24
Add autoregression logic for Flood ML model

tatianawu:


usl_models/usl_models/flood_ml/model.py
```diff
@@ -118,12 +173,82 @@ def forward(
...
+                - "temporal" (required): [T]
+                - "spatiotemporal" (optional): a single flood map [H, W]
+
+        Returns:
+            A tuple (all_flood_predictions, max_flood_map) of the flood predictions and
+            the max predicted flood depth at each cell over the entire storm duration.
+            The maps are the same shape as the input: (H, W).
+        """
+        # TODO: Add data validation
+        geo_input = inputs.get("geospatial")
```
waltaskew, line 197: It looks like this will fail below if 'geospatial' isn't present in the dictionary and this returns a None. I'd expect MyPy to return a type errors below because the code doesn't actually support `geo_input` being None. Some thoughts:
can do
```python
inputs["geospatial"]
```
to just go ahead and get the error rather than propagate it forward into unhandled Nones

but if 'geospatial' is in fact a required argument, we can reflect that in the signature for this function 
```python
def call(self, geospatial: Tensor, temporal: Tensor, spatiotemporal: Optional[Tensor]):
```
If you'd rather accept a single input object rather than something like the above, I'd encode the requirements in a data structure rather than a genric dictionary ala
```python
@dataclasses.dataclass(slots=True)
class ModelInputs:
    geospatial: Tensor
    temporal: Tensor
    spatiotemporal: Optional[Tensor] = None

def call(self, inputs: ModelInputs):
```

usl_models/usl_models/flood_ml/model.py
```diff
@@ -118,12 +173,82 @@ def forward(
...
+        """Runs the entire autoregressive model.
+
+        Args:
+            inputs: A dictionary of input tensors.
+                - "geospatial" (required): [H, W, f]
+                - "temporal" (required): [T]
+                - "spatiotemporal" (optional): a single flood map [H, W]
+
+        Returns:
+            A tuple (all_flood_predictions, max_flood_map) of the flood predictions and
```
waltaskew, line 192: Should add the `->` return type annotations for this.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -118,12 +173,82 @@ def forward(
...
+        )
+        # Note: We assume that, if provided, this input is a *single* flood map.
+        # For convenience, we add this to self.flood_maps to align indexing
+        # between flood maps and rainfall ,i.e., the current flooding conditions
+        # and rainfall at time t are stored at index t along the temporal axis.
+        st_input = inputs.get("spatiotemporal")
+        if st_input is None:
+            st_shape = geo_input.shape[:3] + [1]
+            st_input = tf.zeros(st_shape)
+        self.flood_maps.write(0, st_input)
```
waltaskew, line 209: Is that going to overwrite whatever is at `flood_maps[0]`?
Relatedly, I'm wondering if `flood_maps` should be a local variable within the `call` method rather than an instance variable. What happens if you call `call` twice and inherit a flood map from a previous invocation?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -118,12 +173,82 @@ def forward(
...
+
+        # We use 1-indexing for simplicity. Time step t represents the t-th flood
```
waltaskew, line 211: Is it common to use 1-indexing in the land of modelling? It breaks my tiny computer science brain to use anything besides 0-indexing!

usl_models/usl_models/flood_ml/model.py
```diff
@@ -118,12 +173,82 @@ def forward(
...
+        Args:
+            full_time_input: The [T, m] tensor of all rainfall windows. This
+                function retrieves the appropriate windows given the time step.
+            t: Time step in the autoregression.
+
+        Returns:
+            A tuple of tensors for the flood maps and rainfall, having shapes
+            [n, H, W, 1] and [n, m], respectively.
+        """
+        n = min(self.params["n_flood_maps"], t)
```
waltaskew, line 246: Does it represent an error in the caller if they call this with a t > len(flood_maps) ?
I think it's generally better to raise errors than to try and code around bad inputs.

usl_models/usl_models/flood_ml/model.py
tatianawu: The `call` method expects a dict or list of tensors, so I'm not sure how much flexibility we have in defining a data structure; I'd have to test it out.

We do require geospatial and temporal inputs to always be passed. I currently have a TODO to add some data validation in the future, maybe to be done in a preprocessing step, e.g., with feature scaling. But maybe removing the `get` and calling `inputs["geospatial"]` is a sufficient first step here.

usl_models/usl_models/flood_ml/model.py
tatianawu: There shouldn't be anything at `flood_maps[0]`. My logic is that every index should represent the flood conditions at that time step.

Good point on the multiple `call` invocations. I think this also applies to the geospatial CNN outputs too, and both should be local within a single model call. I'll need a bit more time to double check what should be defined at initialization versus specific runs.

usl_models/usl_models/flood_ml/model.py
tatianawu: This isn't a design choice because of ML practices or anything (I think zero-indexing is probably the default still). I mostly did this for better interpretability. With 1-indexing, time t represents the t-th prediction (in layman terms). It also makes indexing into our tensors a bit more consistent; otherwise the `_update_temporal_inputs` method ends up having some tensors indexed at `t` and some at `t+1`, which I found unnecessarily confusing.

usl_models/usl_models/flood_ml/model.py
tatianawu: This is to catch the case when we've made fewer than `n_flood_maps` predictions; in this case, we want to use all the flood maps available to us. Otherwise we would need to pad some zero flood maps at the beginning of this sequence.

Another way of looking at this is: The model looks at the most recent flood conditions to predict the flooding at the next time step. If we only have a few flood maps, we use all of them. If we have too many, we cap the number to the most recent `n_flood_maps` ones.

usl_models/usl_models/flood_ml/model.py
waltaskew: Ah, I haven't used TensorFlow in quite some time, some I'm not familiar with the class-based `call` interface. It looks kinda sk-learn-y?
So in that case, I'd just add something like
```python
if 'geospatial' not in inputs:
    raise ValueError('Missing required tensor "geospatial"')
```
and ignore everything else I wrote :)

usl_models/usl_models/flood_ml/model.py
tatianawu: Done.

usl_models/usl_models/flood_ml/model.py
tatianawu: Made `flood_maps` a local variable. Also realized that I need to re-set `flood_maps` to the output of the write, since this isn't an in-place operation.

For now, I don't want to worry about whether to pass the `geo_cnn_output` around between various functions and storing it properly, so I'm clearing the output at the end of the model call and setting it back to None. Maybe not the cleanest approach, but I'm still thinking about whether I want to structure the model differently, and I don't want to overcomplicate this PR.

usl_models/usl_models/flood_ml/model.py
tatianawu: Done. Pulled this out into a separate preprocessing function.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -23,65 +30,111 @@ class FloodConvLSTM(tf.keras.Model):
...
+                    (conv_lstm_height, conv_lstm_width, self.params["lstm_units"])
+                ),
                 layers.Conv2DTranspose(8, 4, strides=4, **output_cnn_params),
                 layers.Conv2DTranspose(1, 1, strides=1, **output_cnn_params),
-            ]
+            ],
+            name="output_cnn",
         )
 
     def forward(
```
SallyElHajjar, line 133: we are considering all the data, so the next step will be to include validation and test?

SallyElHajjar (approved):

usl_models/usl_models/flood_ml/model.py
tatianawu: Everything will be done incrementally. This PR just finishes implementing the model logic so that it's actually callable. The actual pipeline work (training, evaluation) will come later, likely also via separate PRs.

SallyElHajjar: Thank you!

PR #39, opened 2024-05-08
Study area input files preparation.

rsutormin:


usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
```diff
@@ -0,0 +1,148 @@
...
+
+def transform_study_area_files(
+    output_dir_path: pathlib.Path | str,
+    elevation_file_path: pathlib.Path | str,
+    sub_area_boundaries_shape_file_path: Optional[pathlib.Path | str] = None,
+    buildings_shape_file_path: Optional[pathlib.Path | str] = None,
+    green_areas_shape_file_path: Optional[pathlib.Path | str] = None,
+    soil_classes_shape_file_path: Optional[pathlib.Path | str] = None,
+    soil_classes_shape_file_property: Optional[str] = None,
+    elevation_output_file_name: str = file_names.ELEVATION_TIF,
```
waltaskew, line 49: One thing to consider is passing these file handles rather than output file names. In the actual use case, we may be writing these to GCS rather than to a local disk. So it'd be nice to be able to call this with
```python
with open('local/path/to/elvevation.tiff',  'wb') as fd:
  transform_study_area_files(elevation_output_file=gd)
```
or
```python
with buck.blob('gcs/path/to/elvevation.tiff').open('wb') as fd:
    transform_study_area_files(elevation_output_file=gd)
```

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
```diff
@@ -0,0 +1,148 @@
...
+        ).header
+    crs = elevation_header.crs.to_string()
+
+    output_dir = pathlib.Path(output_dir_path)
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+
+    sub_area_bbox: Optional[geo_data.BoundingBox] = None
+    if sub_area_boundaries_shape_file_path is not None:
+        print("Preparing Boundaries...")
```
waltaskew, line 95: Looks like a debug line here & a few spots below?
Or did you intend for this to be user output when run via a CLI?
I think it's usually better to use logging facilities `logging.info("Preparing Boundaries...")` from library functions, as the binary's `main` can configure where the logging output should go (stdout, a file, etc.)

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
rsutormin: That's good idea, and I was thinking about it too... But, the problem here is that we don't know which files are present and which are not (based on user's config). Will we be able to pass optional file handlers (depending on the input files user has)? I guess if we go the route you propose, the caller might need to specify some provider opening new file writer to store to on demand.

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
```diff
@@ -0,0 +1,148 @@
...
+        mask_value_feature_property=mask_value_feature_property,
+    )
+    filtered_polygons = (
+        polygons
+        if sub_area_bbox is None
+        else list(
+            polygon_transformers.crop_polygons_to_sub_area(polygons, sub_area_bbox)
+        )
+    )
+    with pathlib.Path(output_polygon_file_path).open("w") as output_polygon_file:
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
```diff
@@ -0,0 +1,148 @@
...
+from usl_lib.writers import polygon_writers
+
+
+def _transform_shape_file(
+    input_shape_file_path: pathlib.Path | str,
+    output_polygon_file_path: pathlib.Path | str,
+    sub_area_bbox: Optional[geo_data.BoundingBox],
+    target_crs: str,
+    mask_value_feature_property: Optional[str] = None,
+) -> None:
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
rsutormin: Also I need some files to be local for further transformations (elevation.tif -> esri ascii, soil classes -> green areas).

rsutormin: Oh, and I forgot about another reason: I see that you upload elevation file to several buckets (one is for metastore, another for chunking in form of TAR) and I guess other input files will go to several places too?

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
rsutormin: Done

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
waltaskew: > Oh, and I forgot about another reason: I see that you upload elevation file to several buckets (one is for metastore, another for chunking in form of TAR) and I guess other input files will go to several places too?

I'm only uploading the raw elevation file to one bucket (the study-areas bucket)
It gets written as chunks to another bucket.

> That's good idea, and I was thinking about it too... But, the problem here is that we don't know which files are present and which are not (based on user's config). Will we be able to pass optional file handlers (depending on the input files user has)? I guess if we go the route you propose, the caller might need to specify some provider opening new file writer to store to on demand.

Optional file handlers are fine -- can just pass a `None` in here and avoid writing to it here. It'd be up to the caller to either pass a writable file handler or `None` here.

> Also I need some files to be local for further transformations (elevation.tif -> esri ascii, soil classes -> green areas).

That is a good reason!

waltaskew (approved):

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
rsutormin: I decided to decompose this giant function into smaller parts that can be directly implemented in the uploader main.py (without passing cloud storage file handlers into other functions from abstract layer).

waltaskew (approved):

PR #40, opened 2024-05-09
Preparation of green areas for CityCat based on soil classes data

rsutormin:


usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Returns:
+        Corrected soil classes that can be used as green areas.
+    """
+    green_areas_raster = polygon_transformers.rasterize_polygons(
+        elevation_header, green_areas_polygons
+    )
+    soil_classes_raster = polygon_transformers.rasterize_polygons(
+        elevation_header, soil_classes_polygons
+    )
```
waltaskew, line 34: If we have the rasters, I'd expect that we could use the green areas as a mask with something like
```python
soil_classes_raster[green_areas_raster == 0] = 0
```
to set the soil type of non-green areas (`green_areas_raster == 0`) to 0

rsutormin: We could do that but how exactly would it help with this logic?

waltaskew (approved):

PR #41, opened 2024-05-10
Add Terraform Config for Feature Matrix Creation Cloud Function

waltaskew:
- Add script cp_cloud_function_files use to prepare the .zip file containing cloud function source code.
- Add terraform config for a cloud function which builds feature tensors for chunks on writes to the chunk bucket.

### DIFF
--- a/terraform/cp_cloud_function_files.sh
+++ b/terraform/cp_cloud_function_files.sh
@@ -1,0 +1,6 @@
+set -e
+
+mkdir -p "$(dirname $0)/files/cloud_function_source"
+rsync "$(dirname $0)/../usl_pipeline/cloud_functions/main.py" "$(dirname $0)/files/cloud_function_source/"
+rsync "$(dirname $0)/../usl_pipeline/cloud_functions/requirements.txt" "$(dirname $0)/files/cloud_function_source/"
+rsync -r --include '*.py' --exclude '*.pyc' "$(dirname $0)/../usl_pipeline/usl_lib/usl_lib" "$(dirname $0)/files/cloud_function_source/"
--- a/terraform/generate_feature_matrix.tf
+++ b/terraform/generate_feature_matrix.tf
@@ -1,0 +1,150 @@
+data "google_project" "project" {
+}
+
+# Place the source code for the cloud function into a GCS bucket.
+data "archive_file" "source" {
+  type        = "zip"
+  source_dir  = "${path.module}/files/cloud_function_source"
+  output_path = "${path.module}/files/cloud_function_source.zip"
+}
+
+resource "google_storage_bucket" "source" {
+  name     = "climateiq-cloud-functions"
+  location = "us-west1"
+}
+
+resource "google_storage_bucket_object" "source" {
+  name   = basename(data.archive_file.source.output_path)
+  bucket = google_storage_bucket.source.name
+  source = data.archive_file.source.output_path
+}
+
+resource "google_storage_bucket" "chunks" {
+  name     = "climateiq-study-area-chunks"
+  location = "us-west1"
+}
+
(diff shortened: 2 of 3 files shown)


terraform/cp_cloud_function_files.sh
```diff
@@ -0,0 +1,6 @@
+set -e
```
waltaskew, line 1: Do folks have thoughts on how to integrate this more with the terraform? Right now, my process is to do
```
./cp_cloud_function_files.sh && terraform apply
```
to deploy new code to the cloud functions. It'd be rad if it was just `terraform apply`

alevenberg:
Have you tested apply/plan/destroy on a project with this config? If so, do you have instructions for what commands you use?

waltaskew:
> Have you tested apply/plan/destroy on a project with this config? If so, do you have instructions for what commands you use?

commands are
`./cp_cloud_function_files.sh && terraform apply`
I left a comment in here -- curious if folks have suggestions for folding in that shell script so it's just a `terraform apply`

the infrastructure described in here is up and running!

alevenberg:
> > Have you tested apply/plan/destroy on a project with this config? If so, do you have instructions for what commands you use?
> 
> commands are `./cp_cloud_function_files.sh && terraform apply` I left a comment in here -- curious if folks have suggestions for folding in that shell script so it's just a `terraform apply`
> 
> the infrastructure described in here is up and running!

It seems strange to me that you would need both. I think you are copying files into a remote destination (bucket). You can upload a file in terraform or point to a docker image for the code (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions_function) (i think)

waltaskew:
> > > Have you tested apply/plan/destroy on a project with this config? If so, do you have instructions for what commands you use?
> > 
> > 
> > commands are `./cp_cloud_function_files.sh && terraform apply` I left a comment in here -- curious if folks have suggestions for folding in that shell script so it's just a `terraform apply`
> > the infrastructure described in here is up and running!
> 
> It seems strange to me that you would need both. I think you are copying files into a remote destination (bucket). You can upload a file in terraform or point to a docker image for the code (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions_function) (i think)

`cp_cloud_function_files.sh` does not copy files into GCS. It moves files into a local cloud_function_source directory. From there, teraform is responsible for zipping the contents of that directory and uploading it to GCS.

terraform/cp_cloud_function_files.sh
alevenberg: I wonder if you can use `repo_source` somehow. I don't know much about cloud functions though. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions2_function#repo_source

prashantkul (approved):

terraform/cp_cloud_function_files.sh
waltaskew: We aren't working with Cloud Source Repositories, so I don't think we can use `repo_source`

alevenberg: I'm fine with you submitting as is-- I don't know much about cloud functions, but it seems strange that we have to run a script before `terraform apply`. But we can change this later on too.

waltaskew: > I'm fine with you submitting as is-- I don't know much about cloud functions, but it seems strange that we have to run a script before `terraform apply`. But we can change this later on too.

I specifically asked if folks had ideas on how to get rid of that `cp_cloud_function_files.sh` script and incorporate it into terraform!

`data "archive_file"` is a bit restrictive. The script is just copying files to make them easier for terraform to pull into a zip via `data "archive_file"`

PR #42, opened 2024-05-13
Only Write Chunk Errors From Chunk Cloud Function

waltaskew:
- Only attempt to write feature chunk errors to the metastore from the feature chunk cloud function. Accomplish this by adding an argument to the `retry_and_report_errors` decorator for additional custom error handling and adding the additional metastore write when decorating the feature chunk cloud function.

### DIFF
--- a/terraform/flood_config_functions.tf
+++ b/terraform/flood_config_functions.tf
@@ -1,0 +1,150 @@
+# Create buckets for storing raw files, file chunks and processed feature matrix chunks.
+resource "google_storage_bucket" "city_cat_config" {
+  name     = "climateiq-citycat-config"
+  location = "us-west1"
+}
+
+# Create a service account used by the function and Eventarc trigger
+resource "google_service_account" "city_cat_config" {
+  account_id   = "gcf-city-cat-sa"
+  display_name = "record-citycat-config cloud function service account"
+}
+
+resource "google_project_iam_member" "city_cat_config_invoking" {
+  project    = data.google_project.project.project_id
+  role       = "roles/run.invoker"
+  member     = "serviceAccount:${google_service_account.city_cat_config.email}"
+  depends_on = [google_project_iam_member.gcs_pubsub_publishing]
+}
+
+resource "google_project_iam_member" "city_cat_config_event_receiving" {
+  project    = data.google_project.project.project_id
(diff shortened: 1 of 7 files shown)


rsutormin (approved):
I see that you're not testing this code, right? I guess it's something not very easy to do?

waltaskew:
> I see that you're not testing this code, right? I guess it's something not very easy to do?

The change here is that `write_study_area_metadata` no longer calls `_write_chunk_metastore_error`. All the other error reporting stuff is staying the same, and I already have tests covering it. I didn't add a test ensuring that the `_write_chunk_metastore_error` call doesn't happen because it's a little awkward to assert that a thing doesn't happen, but I can do so if you think that's useful.

rsutormin (approved):
Well, when I see some complex logic that you change in prod code and then you don't have to also touch test code to compensate the change, my instincts tell me that there is poor test coverage. :)
In most cases I vote for extracting some complex logic from top level main class/file to avoid exactly the situation when it's not super convenient to add tests.
In this particular case I rely on your understanding of what test coverage is enough. Just general thing that I felt I need to flag.

waltaskew:
> Well, when I see some complex logic that you change in prod code and then you don't have to also touch test code to compensate the change, my instincts tell me that there is poor test coverage. :) In most cases I vote for extracting some complex logic from top level main class/file to avoid exactly the situation when it's not super convenient to add tests. In this particular case I rely on your understanding of what test coverage is enough. Just general thing that I felt I need to flag.

This PR keeps existing error reporting & retry logic the same. The tests are helpfully ensuring that the behavior stays the same. The tests are in here:
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/70c827e3947ea61904df2913a435c021bc20e661/usl_pipeline/cloud_functions/main_test.py#L127
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/70c827e3947ea61904df2913a435c021bc20e661/usl_pipeline/cloud_functions/main_test.py#L186

The only change is that `_write_chunk_metastore_error` is only called when `build_feature_matrix` errors, not when `write_study_area_metadata` errors. So the test I could add is a test ensuring that `write_study_area_metadata` does not attempt to call `write_study_area_metadata` when it errors.

PR #46, opened 2024-05-16
Write Un-Padded Number of Rainfall Entries to Metastore

waltaskew:
- Record the un-padded number of timesteps in the rainfall vector. Model training needs access to this so it knows how many timesteps to train against in the auto-regressive loop.

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -174,7 +174,7 @@ def write_flood_scenario_metadata_and_features(
             cloud_event.data["name"]
         )
         with config_blob.open("rt") as rain_fd:
-            as_vector = _rain_config_as_vector(rain_fd)
+            as_vector, length = _rain_config_as_vector(rain_fd)
 
         vector_blob = storage_client.bucket(cloud_storage.FEATURE_CHUNKS_BUCKET).blob(
             f"rainfall/{file_name.with_suffix('.npy')}"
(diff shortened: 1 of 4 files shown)


waltaskew:
@rsutormin friendly ping

usl_pipeline/cloud_functions/main.py
```diff
@@ -184,25 +184,31 @@ def write_flood_scenario_metadata_and_features(
...
+      the rainfall pattern as a numpy array, padded to an agreed-upon
+      length, and `length` represent the un-padded number entries in
+      the array (i.e. the length the array would be if it was not
+      padded.)
+
     """
     entries = config_readers.read_rainfall_amounts(rain_fd)
     return numpy.pad(
         numpy.array(entries, dtype=numpy.float32),
         (0, _RAINFALL_VECTOR_LENGTH - len(entries)),
```
rsutormin, line 210: Do we need any protection in case there are more than _RAINFALL_VECTOR_LENGTH entries in the config file?

rsutormin (approved):

usl_pipeline/cloud_functions/main.py
waltaskew: That's a good point! I can add a defense around this in the config uploader over here, which is also how we ensure things like entries being five minutes apart:
https://github.com/UrbanSystemsLab/climateiq-flood-model/blob/75797dcf9dd83860e8930d16ccc17c9556eae13c/config_uploader/config_uploader/main.py#L130

waltaskew: Added a check in here.

PR #47, opened 2024-05-17
Export elevation from GeoTIFF to Esri ASCII format for CityCat

rsutormin:


usl_pipeline/usl_lib/usl_lib/transformers/elevation_transformers.py
```diff
@@ -0,0 +1,61 @@
...
+            src, no_data_value
+        )
+        elevation_writers.write_header_to_esri_ascii_raster_file(header, output_file)
+
+        if boundaries_polygons is not None:
+            boundaries_raster = polygon_transformers.rasterize_polygons(
+                header, boundaries_polygons
+            )
+
+        data = src.read(band)
```
waltaskew, line 46: When I check the type of `data`, I get a normal `<class 'numpy.ndarray'>`
so it looks like this reads the whole tiff into memory.
What were you planning to call `transform_geotiff_with_boundaries_to_esri_ascii` on? The whole elevation file? Or just the chunks? If just the chunks, do you have a plan for how to pull the chunks into a single esri file for citycat?

usl_pipeline/usl_lib/usl_lib/transformers/elevation_transformers.py
```diff
@@ -0,0 +1,61 @@
...
+        row_index = 0
+        for row in data:
+            if no_data_value is not None:
+                row[row == input_nodata] = no_data_value
```
waltaskew, line 50: It should be much more performant to run this as a vector operation over the full array rather than row by row
```python
data[data == input_nodata] = no_data_value]
```

usl_pipeline/usl_lib/usl_lib/transformers/elevation_transformers.py
```diff
@@ -0,0 +1,61 @@
...
+            row_values = row.tolist()
+            # Zero-out elevation cells where boundaries raster is 0 (meaning outside the
+            # boundaries)
+            if boundaries_raster is not None:
+                boundaries_raster_row_values = boundaries_raster[row_index].tolist()
+                for col_index in range(len(row_values)):
```
waltaskew, line 56: This loop looks equivalent to
```python
row[boundaries_raster[row_index] == 0] = no_data_value
```
or on the full dataset
```python
data[boundaries_raster == 0] = no_data_value
```
which will be actual orders of magnitude more performant than the loops. Vector operations in numpy are much much more performant than Python loops. Looping over a numpy array is something to always try to avoid.

waltaskew: some numbers on a simple example
```python
import numpy
import timeit


def loops(arr):
    for i, row in enumerate(arr):
        for j, x in enumerate(row):
            arr[i, j] = x + 1


def vectors(arr):
    return arr + 1


for shape in ((500, 500), (5000, 5000), (10000, 10000), (15000, 15000)):
    arr = numpy.random.rand(*shape)
    number = 3
    print(
        f"loops {shape}",
        timeit.timeit(lambda: loops(arr), setup="pass", number=number) / number,
    )
    print(
        f"vectors {shape}",
        timeit.timeit(lambda: vectors(arr), setup="pass", number=number) / number,
    )
```
```
loops (500, 500) 0.026520140003412962
vectors (500, 500) 0.0004385486730219175
loops (5000, 5000) 2.7367708566598594
vectors (5000, 5000) 0.056308782659471035
loops (10000, 10000) 11.010823343327502
vectors (10000, 10000) 0.29148437933569465
loops (15000, 15000) 26.96899208734006
vectors (15000, 15000) 0.7613911676647452
```

rsutormin: Yep, I knew about numpy being more efficient. Just realized that rasterio reads the whole thing into memory as numpy. Rewrote that to what you proposed, but with chunking.

usl_pipeline/usl_lib/usl_lib/transformers/elevation_transformers.py
rsutormin: Rewrote that with chunking over rows. And since it's GDAL-based, I moved the code to uploader module.

usl_pipeline/usl_lib/usl_lib/transformers/elevation_transformers.py
rsutormin: Done.

waltaskew (approved):
LGTM -- that's a great approach for avoiding reading the whole file in!

PR #48, opened 2024-05-17
Allow FloodModel to accept multiple storm durations

tatianawu:
The model will now accept a list of inputs, grouped by storm duration, and will fit on them incrementally.

Data are now expected to be structured as FloodModelData objects, which provides better validation.

### DIFF
--- a/usl_models/usl_models/flood_ml/model.py
+++ b/usl_models/usl_models/flood_ml/model.py
@@ -1,5 +1,6 @@
 """Flood model definition."""
 
+import dataclasses
 from typing import TypeAlias
 
 import tensorflow as tf
@@ -12,6 +13,7 @@ from usl_models.flood_ml import model_params
 st_tensor: TypeAlias = tf.Tensor
 geo_tensor: TypeAlias = tf.Tensor
 temp_tensor: TypeAlias = tf.Tensor
+FloodModelData: TypeAlias = data_utils.FloodModelData
 FloodModelParams: TypeAlias = model_params.FloodModelParams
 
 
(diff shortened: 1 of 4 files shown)


SallyElHajjar (approved):

usl_models/usl_models/flood_ml/model.py
```diff
@@ -1,5 +1,6 @@
 """Flood model definition."""
 
+from dataclasses import replace
```
waltaskew, line 3: our [style guide sez](https://google.github.io/styleguide/pyguide.html#22-imports) to only import modules rather than individual classes or functions

usl_models/usl_models/flood_ml/model.py
```diff
@@ -48,72 +50,102 @@ def _build_model(self) -> tf.keras.Model:
...
 
         Returns:
-            A dictionary of processed input tensors.
+            A processed FloodModelData object.
         """
-        if "geospatial" not in inputs:
-            raise ValueError("Missing required tensor 'geospatial'.")
-        if "temporal" not in inputs:
-            raise ValueError("Missing required tensor 'temporal'.")
+        if training:
```
waltaskew, line 68: looks like
```python
if training and data.labels is None:
    raise ValueError("Labels must be provided during model training.")
```

usl_models/usl_models/flood_ml/model.py
```diff
@@ -48,72 +50,102 @@ def _build_model(self) -> tf.keras.Model:
...
         return history
 
+    def train(
+        self,
+        data: list[FloodModelData],
+    ) -> list[tf.keras.callbacks.History]:
+        """Trains the model.
+
+        Args:
+            data: A list of FloodModelData objects. Objects should be grouped by storm
```
waltaskew, line 134: 'grouped by storm duration' in what sense? Partly a data structure questions -- should it be
```python
    storm_duration: int
    geospatial: Iterable[tf.Tensor]
```
to group all tensors with the same storm duration?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -48,72 +50,102 @@ def _build_model(self) -> tf.keras.Model:
...
-            raise ValueError("Missing required tensor 'geospatial'.")
-        if "temporal" not in inputs:
-            raise ValueError("Missing required tensor 'temporal'.")
+        if training:
+            if data.labels is None:
+                raise ValueError("Labels must be provided during model training.")
+
+            expected_label_shape = list(data.labels.shape)
+            expected_label_shape[1] = data.storm_duration
+            assert data.labels.shape == expected_label_shape, (
```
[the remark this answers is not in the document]
tatianawu: Yup. I guess this is a bit more clear,  will make the switch.

usl_models/usl_models/flood_ml/model.py
tatianawu: Ack, will fix.

I use `from dataclasses import dataclass` in several places. Is this an exception, or should I change those too?

usl_models/usl_models/flood_ml/model.py
tatianawu: Yes. But there was another block to the logic for checking the expected labels. Would you recommend pulling that into its own `if` block?

usl_models/usl_models/flood_ml/model.py
tatianawu: This is a poor description on my part. Each `FloodModelData` instance is associated with a single storm duration. The list is to allow passing in data for multiple storm durations. In other words, data for different storm durations must be passed in as different `FloodModelData` objects.

usl_models/usl_models/flood_ml/model.py
waltaskew: The [only exceptions](https://google.github.io/styleguide/pyguide.html#2241-exemptions) are for the typing & collections modules
but I dunno, maybe we can have an honorary one for dataclasses

usl_models/usl_models/flood_ml/model.py
waltaskew: Ah, I missed that! Fine to keep as is. Could also consider
```python
if training and data.labels is None:
    raise ValueError("Labels must be provided during model training.")
elif training:
   ...
```
but I think both are readable (despite me misreading it :)

usl_models/usl_models/flood_ml/model.py
waltaskew: Okay, cool, what you describe here in the comment box sounds good -- might make the language in the docstring as well.

waltaskew (approved):

PR #50, opened 2024-05-22
Add Reader for CityCAT Simulation Files

waltaskew:
- Add a reader for CityCAT rsl output files which returns their flood predictions as a raster.
- Factor out forward_transform & back_transform methods for building affines from elevation headers.


rsutormin (approved):
Looks perfect!

PR #51, opened 2024-05-22
Minor uploader logic correction and logging for uploader top-level tasks

rsutormin:


usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
```diff
@@ -22,32 +24,54 @@ def transform_soil_classes_as_green_areas(
         elevation_header: The source of coordinate system and study area region info.
         green_areas_polygons: List of tuples with green area polygons.
         soil_classes_polygons: List of tuples with soil class polygons.
+        log_details: Indicates that details of intermediate steps should be logged.
 
     Returns:
         Corrected soil classes that can be used as green areas.
     """
+    if log_details:
```
waltaskew, line 32: Rather than have branches for logging statements, what we can do is this:
```python
def main():
  args = parser.parse_args()
  if args.verbose:
    logging.basicConfig(level=logging.INFO)
  else:
    logging.basicConfig(level=logging.ERROR)
    
def foo():
  logging.info('whatever')
  logging.error('oh no!')
```
so the `logging.basicConfig(level=LEVEL)` call determines what level of logging calls are actually emitted and each function can just make `logging.level()` calls without needing to pass `log_details` arguments everywhere.

waltaskew (changes requested):

usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
rsutormin: Yeah, that's what I was looking for! Done.

usl_pipeline/study_area_uploader/study_area_uploader/main.py
```diff
@@ -153,5 +158,12 @@ def parse_args() -> argparse.Namespace:
         help="Band index in GeoTIFF file containing elevation data (default value is"
         + " 1)",
     )
+    parser.add_argument(
+        "--verbose",
+        type=bool,
```
waltaskew, line 163: I think you'd need to do something like
```python
parser.add_argument('--verbose', action=argparse.BooleanOptionalAction)
```
As is, I think it'd treat this like `--verbose argument_you_passed_into_verbose_arg`

waltaskew (approved):
Does that logic correction deserve a test?

usl_pipeline/study_area_uploader/study_area_uploader/main.py
rsutormin: Yep! Done.

PR #52, opened 2024-05-27
Extend build_feature_matrix_from_archive to handle heat modeling files

saskatchuwan:
Add initial logic and unit test for modifying `build_feature_matrix` to handle wps files for AtmoML. Follow-up PRs will address:
- setting up the appropriate matrix transforms on the WPS variables that are in-scope for ML training
- writing to metastore

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -13,6 +13,7 @@ from google.cloud import storage
 import functions_framework
 import numpy
 from numpy.typing import NDArray
+import xarray
 import rasterio
 
 from usl_lib.readers import config_readers
@@ -289,13 +290,29 @@ def _build_feature_matrix_from_archive(
                 continue
 
             name = pathlib.PurePosixPath(member.name).name
+            # TODO: Group logic branch by path (per hazard model)
             if name == file_names.ELEVATION_TIF:
                 return _read_elevation_features(fd)
+            # Handle heat model input files (WPS)
+            elif name.startswith("met_em") and name.endswith(".nc"):
+                return _read_wps_features(fd)
             # TODO: handle additional archive members.
             else:
                 logging.warning(f"Unexpected member name: {name}")
 
     return None, FeatureMetadata()
+
+
+def _read_wps_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
+    # Ignore type checker error - BytesIO inherits from expected type BufferedIOBase
+    # https://shorturl.at/lk4om
+    with xarray.open_dataset(fd) as ds:  # type: ignore
+        feature = ds.data_vars["SNOALB"].values
+        # Drop time axis - each WPS output file will correspond to single time value
+        feature = numpy.squeeze(feature, axis=0)
+
+    # TODO: Write to metastore
+    return feature, FeatureMetadata()
 
 
 def _read_elevation_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
--- a/usl_pipeline/cloud_functions/main_test.py
+++ b/usl_pipeline/cloud_functions/main_test.py
@@ -7,6 +7,7 @@ from unittest import mock
 from google.cloud import firestore
 from google.cloud import storage
 import functions_framework
+import netCDF4
 import numpy
 import rasterio
 
@@ -15,7 +16,7 @@ import main
 
 @mock.patch.object(main.firestore, "Client", autospec=True)
 @mock.patch.object(main.storage, "Client", autospec=True)
-def test_build_feature_matrix(mock_storage_client, mock_firestore_client):
+def test_build_feature_matrix_flood(mock_storage_client, mock_firestore_client):
     # Get some random data to place in a tiff file.
     height = 2
     width = 3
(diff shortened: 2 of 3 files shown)


usl_pipeline/cloud_functions/main_test.py
```diff
@@ -122,6 +126,89 @@ def test_build_feature_matrix(mock_storage_client, mock_firestore_client):
...
+        lat[:] = [200, 200, 200]
+        lon[:] = [300, 300, 300]
+        sno_alb[:] = numpy.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]])
+
+    # Check the dataset written to the local ncfile
+    with xr.open_dataset("met_em_test.nc") as ds:
+        print("Checking contents of saved ncfile")
+        print(ds)
+
+    # Place the ncfile bytes into an archive.
```
saskatchuwan, line 153: @waltaskew @rsutormin - Could you help take a look at how the unit test is set up? Currently, the unit test is failing when the `main.py:line306`. The netcdf file that was created in the test for some reason can't be loaded as an xarray dataset (error similar to: https://stackoverflow.com/questions/74453978/xarray-error-valueerror-did-not-find-a-match-in-any-of-xarrays-currently-ins), which points to either an incompatible netcdf file format or the file does not exist.

I manually tested the new code with an existing wps file we have on hand and was able to verify that the code worked e2e (reading archive -> uploading npy file). So not sure what i'm doing incorrectly in the unit test here.

Thanks!

waltaskew: I'd wager it's because you aren't setting the size on the TarInfo object. Elsewhere in the file, we have
```
        tf = tarfile.TarInfo("elevation.tif")
        tf.size = len(tiff_bytes)
        tar.addfile(tf, io.BytesIO(tiff_bytes))
```
so without the size, I'd guess you're adding a size zero entry.

saskatchuwan:
@waltaskew How do I fix the test run in the CI pipeline? I think it's complainign about the missing gcloud creds. The test suite runs successfully locally.

waltaskew:
> @waltaskew How do I fix the test run in the CI pipeline? I think it's complainign about the missing gcloud creds. The test suite runs successfully locally.

It looks like it's failing to missing dependencies:
```
ValueError: found the following matches with the input file in xarray's IO backends: ['h5netcdf']. But their dependencies may not be installed, see:
```

it'll then try to report that error to GCP
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/fc3e8c1d0856e48f0895e0b386af6fc475e46791/usl_pipeline/cloud_functions/main.py#L107
where it's presumably missing credentials

You should mock the error reporter so it doesn't actually hit GCP, or else test code paths not wrapped by `_retry_and_report_errors`
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/fc3e8c1d0856e48f0895e0b386af6fc475e46791/usl_pipeline/cloud_functions/main.py#L43

saskatchuwan:
> > @waltaskew How do I fix the test run in the CI pipeline? I think it's complainign about the missing gcloud creds. The test suite runs successfully locally.
> 
> It looks like it's failing to missing dependencies:
> 
> ```
> ValueError: found the following matches with the input file in xarray's IO backends: ['h5netcdf']. But their dependencies may not be installed, see:
> ```
> 
> it'll then try to report that error to GCP
> 
> https://github.com/UrbanSystemsLab/climateiq-cnn/blob/fc3e8c1d0856e48f0895e0b386af6fc475e46791/usl_pipeline/cloud_functions/main.py#L107
> 
> 
> where it's presumably missing credentials
> You should mock the error reporter so it doesn't actually hit GCP, or else test code paths not wrapped by `_retry_and_report_errors`
> 
> https://github.com/UrbanSystemsLab/climateiq-cnn/blob/fc3e8c1d0856e48f0895e0b386af6fc475e46791/usl_pipeline/cloud_functions/main.py#L43

Done and done - thanks Walt. Build is passing now

usl_pipeline/cloud_functions/main.py
```diff
@@ -289,15 +291,51 @@ def _build_feature_matrix_from_archive(
...
+        feature = numpy.squeeze(feature, axis=0)
+        feature = _reorder_wps_spatial_axes(feature)
+        # TODO: Add lat/lon spatial coordinates to array
+
+    # TODO: Write to metastore
+    return feature, FeatureMetadata()
+
+
+def _reorder_wps_spatial_axes(feature: NDArray) -> NDArray:
+    """Reorders the axes for x-lat and y-lon.
```
waltaskew, line 322: Can you describe this a bit more? What are the original axes and how do they need to be re-ordered? (and maybe what is AtmoML and x-lat & y-long and what's their relevance?)
It looks like we're taking a raster of shape (width, height) and returning one of (height, width) -- or maybe the other way around?

usl_pipeline/cloud_functions/main.py
```diff
@@ -289,15 +291,51 @@ def _build_feature_matrix_from_archive(
...
+    Reorders dimensions within the array to align with the order expected from AtmoML.
+
+    Args:
+      feature: The array containing the feature and its dimensions
+
+    Returns:
+      A new array with its spatial dimension axes reordered.
+    """
+    # TODO: indices of lat lon might not always be the same if there are
+    # additional spatial dims. May also need to read in coords to find out what
```
waltaskew, line 333: Are you concerned the different layers of the netcdf won't line up? I'd hope the xarray library returns consistent rasters of each layer.

usl_pipeline/cloud_functions/main_test.py
```diff
@@ -123,6 +124,97 @@ def test_build_feature_matrix(mock_storage_client, mock_firestore_client):
...
+def test_build_feature_matrix_wrf(
+    mock_storage_client,
+    mock_firestore_client,
+    mock_error_reporting_client,
+):
+    # Get some random data to place in a netcdf file
+    netcdf_array = numpy.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]], dtype=numpy.float32)
+
+    # Build a netcdf file
+    with netCDF4.Dataset("met_em_test.nc", mode="w", format="NETCDF4") as ncfile:
```
waltaskew, line 139: It looks like this is writing a file named 'met_em_test.nc' to the file system. Can we use a BytesIO buffer instead? Or at least a tempfile?

usl_pipeline/cloud_functions/main_test.py
```diff
@@ -123,6 +124,97 @@ def test_build_feature_matrix(mock_storage_client, mock_firestore_client):
...
+            mock.call().bucket().blob("study_area/name.npy"),
+        ]
+    )
+
+    mock_archive_blob.open.assert_called_once_with("rb")
+
+    # Ensure we attempted to upload a serialized matrix of the netcdf
+    mock_feature_blob.upload_from_file.assert_called_once_with(mock.ANY)
+    uploaded_array = numpy.load(mock_feature_blob.upload_from_file.call_args[0][0])
+    # TODO: for now, just check for length - update assertion once we finalize the array
```
waltaskew, line 211: Can we assert the transform we're currently performing?
It's fine to update the transform and the tests in the future.

usl_pipeline/cloud_functions/main.py
```diff
@@ -13,6 +13,7 @@
 import functions_framework
 import numpy
 from numpy.typing import NDArray
+import xarray as xr
```
waltaskew, line 16: I think the full `xarray` name is a little clearer. I think folks mostly do `import xarray as xr` or `import numpy as np` or `import pandas as pd` just for brevity in example code (although `np` and `pd` are pretty commonly recognized in scientific computing circles)

usl_pipeline/cloud_functions/main.py
saskatchuwan: Removing this for now - working on refining this in a separate branch

usl_pipeline/cloud_functions/main.py
saskatchuwan: From what I've seen - the axis are sometimes ordered differently, but i will need to some more testing. Removing this for now - working on refining this in a separate branch

usl_pipeline/cloud_functions/main_test.py
saskatchuwan: Thanks - updated this so we're working with an in-memory file

usl_pipeline/cloud_functions/main.py
```diff
@@ -289,15 +290,31 @@ def _build_feature_matrix_from_archive(
...
+                return _read_wps_features(fd)
             # TODO: handle additional archive members.
             else:
                 logging.warning(f"Unexpected member name: {name}")
 
     return None, FeatureMetadata()
 
 
+def _read_wps_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
+    # Ignore type checker error - BytesIO inherits from expected type BufferedIOBase
```
waltaskew, line 307: FYI you can use `typing.cast` to cast this as a BufferedIOBase , which would obviate the `# type: ignore` and also this comment :)
I don't have a strong preference between the two options, just pointing it out if you haven't seen it before.
https://docs.python.org/3/library/typing.html#typing.cast

waltaskew (approved):
LGTM

usl_pipeline/cloud_functions/main.py
saskatchuwan: Tried both but ran into issues..

When I casted into BufferedIOBase, the test fails due to failure to load the dataset with xarray and changing the function signature to expect BufferedIOBase fails the type checker!

PR #54, opened 2024-05-29
Create Terraform Configuration for Study Area Write CF

waltaskew:
- Create terraform configuration for a cloud function which responds to writes to a study area bucket.
- Rename google_service_account.account to google_service_account.generate_feature_matrix as we define several service accounts.


prashantkul (approved):
Looks good!

PR #57, opened 2024-05-29
Exclusion logic for special soil classes that cannot be marked as green areas

rsutormin:
This is a modification of the algorithm that reconstructs (or classify) green areas based on soil classes. It takes into account the requirement to exclude non-green area soil classes.

Before this change algorithm had the following steps:
1) Exclude from soil class polygons all parts of regions that are outside green areas, and report them to CityCat;
2) Classify soil classes for all raster cells from green areas that weren't covered by any soil classes using nearest neighbor approach looking for nearest soil class polygons, and report them as 1-cell polygons to CityCat in addition to item 1.

After the change:
1) Exclude from soil class polygons all parts of regions that are outside green areas, *also filter out polygons with non-green soil classes*, and report them as polygons to CityCat;
2) Classify soil classes for all raster cells from green areas that weren't covered by any soil classes *(including non-green ones)* using nearest neighbor approach looking for nearest *non-green* soil class polygons, and report them as 1-cell polygons to CityCat in addition to item 1.

### DIFF
--- a/usl_pipeline/study_area_uploader/study_area_uploader/main.py
+++ b/usl_pipeline/study_area_uploader/study_area_uploader/main.py
@@ -40,6 +40,7 @@ def main() -> None:
             args.soil_type_mask_feature_property,
             pathlib.Path(temp_dir),
             study_area_bucket,
+            input_non_green_area_soil_classes=set(args.non_green_area_soil_classes),
         )
 
         if args.export_to_citycat:
@@ -166,6 +167,13 @@ def parse_args() -> argparse.Namespace:
         help="Indicates that more details of processing steps should be printed to the"
         + "console (INFO logging level instead of default ERROR one)",
         action=argparse.BooleanOptionalAction,
+    )
+    parser.add_argument(
+        "--non-green-area-soil-classes",
+        type=int,
+        default=[],
+        help="Optional list of soil classes that cannot be treated as green areas",
+        nargs="*",
     )
 
     return parser.parse_args()
--- a/usl_pipeline/study_area_uploader/study_area_uploader/main.py
+++ b/usl_pipeline/study_area_uploader/study_area_uploader/main.py
@@ -176,4 +176,12 @@ def parse_args() -> argparse.Namespace:
         nargs="*",
     )
 
-    return parser.parse_args()
+    args = parser.parse_args()
+
+    # Validation of CLI arguments
+    if args.soil_type_file and not args.non_green_area_soil_classes:
+        parser.error(
+            "--non_green_area_soil_classes required if --soil_type_file present"
+        )
+
+    return args
--- a/usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
+++ b/usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
@@ -40,18 +40,18 @@ def transform_soil_classes_as_green_areas(
         Corrected soil classes that can be used as green areas.
     """
     logging.info("Starting reconstruction of green areas based on soil classes")
-    soil_classes_polygons_non_green_excluded = [
+    green_area_soil_classes_polygons = [
         p for p in soil_classes_polygons if p[1] not in non_green_area_soil_classes
     ]
     logging.info(
         "Intersecting %s soil class polygons with green areas",
-        len(soil_classes_polygons_non_green_excluded),
+        len(green_area_soil_classes_polygons),
     )
     existing_soil_class_polygons = []
     for p in green_areas_polygons:
         existing_soil_class_polygons.extend(
             polygon_transformers.intersect(
-                soil_classes_polygons_non_green_excluded,
+                green_area_soil_classes_polygons,
                 geometry.MultiPolygon([p[0]]),
                 skip_logging=True,
             )
@@ -67,8 +67,9 @@ def transform_soil_classes_as_green_areas(
     )
     logging.info("Soil classes raster is getting created...")
     # Raster for soil classes is made based on soil_classes_polygons that might include
-    # non-green area polygons. This is done specifically so that the cells of those
-    # non-green classes wouldn't be gap-filled.
+    # non-green area polygons. This is to avoid creating cells inside the green spaces
+    # with missing soil values, which would be interpolated, when we need to leave them
+    # as their non-green-area soil type.
     soil_classes_raster = polygon_transformers.rasterize_polygons(
         elevation_header, soil_classes_polygons
     )
@@ -79,7 +80,7 @@ def transform_soil_classes_as_green_areas(
             elevation_header,
             green_areas_raster,
             soil_classes_raster,
-            soil_classes_polygons_non_green_excluded,
+            green_area_soil_classes_polygons,
         )
     )
     logging.info("Green areas reconstruction based on soil classes is done.")
--- a/usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
+++ b/usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
@@ -20,6 +20,15 @@ def transform_soil_classes_as_green_areas(
     raster where soil classes information is missing, the value is assigned from the
     nearest neighbor soil class polygon.
 
+    Here are formal steps of the algorithm:
+      - Exclude from soil class polygons all parts of regions that are outside green
+        areas, also filter out polygons with non-green soil classes, and report them as
+        polygons to the output;
+      - Classify soil classes for all raster cells from green areas that weren't covered
+        by any soil classes (including non-green ones) using nearest neighbor approach
+        looking for nearest non-green soil class polygons, and report them as 1-cell
+        polygons to the output in addition to the previous item.
+
     Args:
         elevation_header: The source of coordinate system and study area region info.
         green_areas_polygons: List of tuples with green area polygons.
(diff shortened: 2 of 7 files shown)


usl_pipeline/study_area_uploader/study_area_uploader/main.py
```diff
@@ -167,5 +168,12 @@ def parse_args() -> argparse.Namespace:
         + "console (INFO logging level instead of default ERROR one)",
         action=argparse.BooleanOptionalAction,
     )
+    parser.add_argument(
+        "--non-green-area-soil-classes",
```
waltaskew, line 172: Is this something callers should pass in? Or is this just something we should hard-code?
Is this actually something that will vary across study areas?
My concern would be that callers would have to supply the same value to this script every time to get correct behaviour, and I'm not sure if it's actually something callers need or want to vary.

usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
```diff
@@ -50,6 +57,9 @@ def transform_soil_classes_as_green_areas(
         elevation_header, green_areas_polygons
     )
     logging.info("Soil classes raster is getting created...")
+    # Raster for soil classes is made based on soil_classes_polygons that might include
+    # non-green area polygons. This is done specifically so that the cells of those
+    # non-green classes wouldn't be gap-filled.
```
waltaskew, line 62: what do you mean by 'gap-filled' here?

usl_pipeline/usl_lib/usl_lib/transformers/polygon_transformers.py
```diff
@@ -127,13 +127,15 @@ def crop_polygons_to_sub_area(
...
     multi_polygon_boundaries: geometry.MultiPolygon,
+    skip_logging: bool = False,
 ) -> Iterable[Tuple[geometry.Polygon, int]]:
     """Intersects polygons associated with integer masks with multi-polygon boundaries.
 
     Args:
         polygon_masks_to_process: Input polygons associated with integer masks that
             intersection process should be applied to.
         multi_polygon_boundaries: Multi-polygon area to apply on the input polygons.
+        skip_logging: Optional indicator that logging should be skipped.
```
waltaskew, line 138: Why do we conditionally want to skip logging?

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
```diff
@@ -17,6 +17,9 @@
 from usl_lib.transformers import polygon_transformers
 from usl_lib.writers import polygon_writers
 
+"""Default soil class value that is recognized a non-green area."""
```
waltaskew, line 20: Looks like a # comment rather than a """string"""
I'm having trouble understanding the comment, though. This is the soil type that indicates a non-green area?

waltaskew (commented):
Can you provide a high-level description for this PR? I'm having a little trouble parsing this.

rsutormin:
> Can you provide a high-level description for this PR? I'm having a little trouble parsing this.

Done

usl_pipeline/study_area_uploader/study_area_uploader/main.py
rsutormin: Yeah, I think soil classes are very region specific. I'm not sure there are unified codes recognized by globally.

usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
rsutormin: I mean filled-in in place of absent data (for raster cells covered by non-green soil classes we should NOT do nearest-neighbor filling-in). Do you have any suggestions for how to formulate it better?

usl_pipeline/usl_lib/usl_lib/transformers/polygon_transformers.py
rsutormin: In this particular case logging shows a lot of repeated progress of the intersection process that prevents you from seeing the rest of the logging lines. But I don't want to remove intersection progress completely though because it's useful in other cases.

waltaskew:
Thanks, this is a very helpful explanation. The explanation of the algorithm after the change is very clear and helpful -- would be great to include in the code, maybe in the commit as well.

> This is a modification of the algorithm that reconstructs (or classify) green areas based on soil classes. It takes into account the requirement to exclude non-green area soil classes.
> 
> Before this change algorithm had the following steps:
> 
> 1. Exclude from soil class polygons all parts of regions that are outside green areas, and report them to CityCat;
> 2. Classify soil classes for all raster cells from green areas that weren't covered by any soil classes using nearest neighbor approach looking for nearest soil class polygons, and report them as 1-cell polygons to CityCat in addition to item 1.
> 
> After the change:
> 
> 1. Exclude from soil class polygons all parts of regions that are outside green areas, _also filter out polygons with non-green soil classes_, and report them as polygons to CityCat;
> 2. Classify soil classes for all raster cells from green areas that weren't covered by any soil classes _(including non-green ones)_ using nearest neighbor approach looking for nearest _non-green_ soil class polygons, and report them as 1-cell polygons to CityCat in addition to item 1.

usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
waltaskew: Maybe, 
"This is to avoid creating cells inside the green spaces with missing soil values, which would be interpolated, when we need to leave them as their non-green-area soil type."
?

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
rsutormin: Fixed

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
```diff
@@ -178,6 +201,15 @@ def prepare_and_upload_study_area_files(
                     mask_value_feature_property=soil_class_mask_feature_property,
                 )
             )
+            if len(input_non_green_area_soil_classes) > 0:
+                soil_classes_polygons = [
+                    _update_mask_if_needed(
```
waltaskew, line 206: Is the idea here to collapse the different non-green-area soil classes into a single non-green-area value? Would it also work to continue passing the set `input_non_green_area_soil_classes` around, or does it need to be a single value?

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
```diff
@@ -80,6 +96,8 @@ def prepare_and_upload_study_area_files(
     soil_class_mask_feature_property: Optional[str],
     work_dir: pathlib.Path,
     study_area_bucket: storage.Bucket,
+    input_non_green_area_soil_classes: set[int] = set(),
+    output_non_green_area_soil_class: int = DEFAULT_NON_GREEN_AREA_SOIL_CLASS,
```
waltaskew, line 100: Why would a caller want to pass a non-default value here? Does it make sense to remove this as an argument?
Would they need to pass that same non-default value into `prepare_and_upload_citycat_input_files` as well?

waltaskew (approved):
The remaining questions I have are things that I think might be helpful to answer with code comments in case future readers have the same questions :)
Possible documentation additions aside, this LGTM

usl_pipeline/study_area_uploader/study_area_uploader/main.py
rsutormin: We can't. Soil classes as a whole is an optional data source.

usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
rsutormin: SG, made a correction.

usl_pipeline/study_area_uploader/study_area_uploader/main.py
waltaskew: Could we make it conditionally required if soil classes are present?

usl_pipeline/study_area_uploader/study_area_uploader/transformers/soil_classes_transformers.py
```diff
@@ -23,21 +24,27 @@ def transform_soil_classes_as_green_areas(
...
         green_areas_polygons: List of tuples with green area polygons.
         soil_classes_polygons: List of tuples with soil class polygons.
+        non_green_area_soil_classes: Optional set of soil class values that should be
+            excluded from green areas.
 
     Returns:
         Corrected soil classes that can be used as green areas.
     """
     logging.info("Starting reconstruction of green areas based on soil classes")
+    soil_classes_polygons_non_green_excluded = [
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
rsutormin: Not for the need to also do the same nearest neighbor procedure for every chunk of the full study area (NYC) in cloud_functions. If we could pass around user-provided parameters to cloud_functions, then I wouldn't substitute non-green soil classes by some default value.

usl_pipeline/study_area_uploader/study_area_uploader/transformers/study_area_transformers.py
rsutormin: They actually don't want to pass a non-default value here. I removed this argument and the corresponding one in the citycat preparation function.

usl_pipeline/study_area_uploader/study_area_uploader/main.py
rsutormin: Done

PR #59, opened 2024-05-31
[wrf] Modify build_feature_matrix to account for multiple feature vars

saskatchuwan:
- Create new `wps_variables.py` file to track ordered list of required variables for ML
- Update `build_feature_matrix` to build features matrix with list of required WPS variables stacked

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -21,6 +21,7 @@ from usl_lib.readers import elevation_readers
 from usl_lib.storage import cloud_storage
 from usl_lib.storage import file_names
 from usl_lib.storage import metastore
+from usl_lib.shared import wps_variables
 
 _MAX_RETRY_SECONDS = 60 * 2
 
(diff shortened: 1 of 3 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -307,12 +308,38 @@ def _read_wps_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
...
+        features_matrix = numpy.array([])
+        for var in wps_variables.REQUIRED_VARS:
+            try:
+                feature = ds.data_vars[var].values
+                feature = _process_wps_feature(feature)
+
+                if not features_matrix.size:
+                    features_matrix = feature
+                else:
+                    features_matrix = numpy.dstack((features_matrix, feature))
```
waltaskew, line 320: might refactor this as
```python
feature_components = []
for var in wps_variables.REQUIRED_VARS:
  feature = ds.data_vars[var].values
  ...
  feature_components.append(feature)

feature_matrix = numpy.dstack(feature_components)
```
I'd wager that a single dstack will be faster than repeated invocations.

usl_pipeline/cloud_functions/main.py
```diff
@@ -307,12 +308,38 @@ def _read_wps_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
...
+            except KeyError:
+                logging.warning(f"No variable named '{var}' found in dataset.")
```
waltaskew, line 322: We'd need to stop the show here rather than log & continue. ML will only be happy if the tensor is of the exact expected dimensions, so we'd just be deferring an error until training or prediction time.

usl_pipeline/cloud_functions/main.py
```diff
@@ -307,12 +308,38 @@ def _read_wps_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
...
     # TODO: Write to metastore
-    return feature, FeatureMetadata()
+    return features_matrix, FeatureMetadata()
+
+
+def _process_wps_feature(feature: NDArray) -> NDArray:
+    """Performs a series of array transforms on a WPS variable array.
+
+    1. Drops time axis - since each WPS output file corresponds to one datetime
+    2. TBD (feature scaling, extracting levels, etc)...
```
waltaskew, line 332: FYI things like scaling will be different per feature type, so we'd need the feature name or such to dispatch behaviour against in here.

saskatchuwan: Good call - updated to pass in the xarray dataarray object instead

usl_pipeline/cloud_functions/main.py
saskatchuwan: Makes sense - remove the try/catch - will let the keyerror stop processing

waltaskew (approved):
LGTM

PR #63, opened 2024-05-31
Some Metastore Refactoring

waltaskew:
- Use 'uri' rather than 'path' for fields containing the full GCS URI
- Allow creating, deleting and retrieving `FloodScenarioConfig`s by ID rather than full URI.
- Add `get_ref` method for FloodScenarioConfig and StudyArea and refactor to use it internally.
- Add `as_header` method to StudyArea objects.

### DIFF
--- a/usl_pipeline/usl_lib/usl_lib/storage/metastore.py
+++ b/usl_pipeline/usl_lib/usl_lib/storage/metastore.py
@@ -3,10 +3,14 @@ from typing import Optional
 import urllib.parse
 
 from google.cloud import firestore
+import rasterio
+
+from usl_lib.shared import geo_data
 
 
 STUDY_AREAS = "study_areas"
-CHUNKS = "chunks"
+STUDY_AREA_CHUNKS = "chunks"
+
 CITY_CAT_RAINFALL_CONFIG = "city_cat_rainfall_configs"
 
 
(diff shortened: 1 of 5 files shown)


usl_pipeline/usl_lib/usl_lib/storage/metastore.py
```diff
@@ -35,21 +39,33 @@ class StudyArea:
...
         """Creates a new study area in the given DB. Only writes non-None attributes.
 
         Args:
           db: The firestore database client to use for the read.
         """
-        db.collection(STUDY_AREAS).document(self.name).create(self._as_dict())
+        self.get_ref(db, self.name).create(self._as_dict())
 
     def set(self, db: firestore.Client) -> None:
         """Creates a new study area in the given DB. Only writes non-None attributes.
```
rsutormin, line 63: I guess it's "Creates or updates" in the doc for "set" method?

rsutormin (approved):

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
waltaskew: `create()` only creates and raises an error if the document already exists
`set()` creates or replaces if it already exists

waltaskew: fixed docstring

PR #65, opened 2024-06-04
Add Simulation to Metastore

waltaskew:
- Add helpers for writing & retrieving simulations and simulation label chunks into the metastore.


rsutormin (approved):
There is one problem with processing configuration via this approach: we need to load `InfiltrationParams` from configuration in order to properly re-scale soil-classes data of feature chunks. But we load pointers to these configs only when simulation results are ready for analysis. The work-around I see here is to postpone green areas raster re-scaling till the feature data is merged with label data for training.

PR #67, opened 2024-06-04
Study area chunker.

rsutormin:
Cuts elevation and polygon data of study area into chunks, groups chunk files into TARs and uploads them to cloud storage bucket.

For elevation data 2 modifications are applied to GeoTIFF file of each chunk:
 - NODATA value is set to default (-9999.0)
 - main band number is set to 1

This is done in order to avoid options for these values during the subsequent processing on the cloud_functions side of the pipeline.


usl_pipeline/study_area_uploader/study_area_uploader/main.py
```diff
@@ -153,6 +165,14 @@ def parse_args() -> argparse.Namespace:
         + " to CityCat storage bucket.",
         action=argparse.BooleanOptionalAction,
     )
+    parser.add_argument(
+        "--upload-chunks",
```
waltaskew, line 169: What's a case where the user wouldn't want to upload chunks? I think we always want to upload them.

usl_pipeline/study_area_uploader/tests/chunkers/test_study_area_chunkers.py
```diff
@@ -0,0 +1,255 @@
...
+def copy_stream_to_file(
+    input_fd: typing.BinaryIO,
+    output_file_path: pathlib.Path,
+) -> None:
+    """Copies file stream to local file (wrapper around shutil.copyfileobj)."""
+    with open(output_file_path, "wb") as output_fd:
+        shutil.copyfileobj(input_fd, output_fd)
+
+
+def mock_file_blob(work_dir: pathlib.Path, relative_path: str) -> storage.Blob:
```
waltaskew, line 68: for testing the
```python
        study_area_chunk_bucket.blob(
            f"{study_area_name}/{chunk_file_name}.tar"
        ).upload_from_file(tar_fd)
```
call, what I'd do is
```python
mock_bucket.blob.assert_has_calls([
  mock.call('study_area_name/chunk_1.tar'),
  mock.call('study_area_name/chunk_2.tar'),
])
tar_uploaded_for_chunk_1 = mock_bucket.blob().upload_from_file.mock_calls[0].args[0]
tar_uploaded_for_chunk_2 = mock_bucket.blob().upload_from_file.mock_calls[1].args[0]
```
and then you can assert properties of the tar you uploaded.

usl_pipeline/study_area_uploader/study_area_uploader/chunkers/study_area_chunkers.py
```diff
@@ -0,0 +1,160 @@
...
+    if chunks_dir is not None:
+        tar.add(str(chunks_dir / chunk_file_name), arcname=tar_entry_file_name)
+
+
+def build_and_upload_chunks(
+    study_area_name: str,
+    input_data: study_area_transformers.PreparedInputData,
+    work_dir: pathlib.Path,
+    study_area_chunk_bucket: storage.Bucket,
+    chunk_size: int = 1000,
```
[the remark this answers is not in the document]
rsutormin: Since I actually do have default value = 1000 in the uploader main.py CLI args (https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_pipeline/study_area_uploader/study_area_uploader/main.py#L145) I don't need to have it hardcoded in the code. On the cloud_function side though I would prefer to load it from metadata instead of having it hardcoded...

usl_pipeline/study_area_uploader/study_area_uploader/main.py
rsutormin: Agreed. Removed it. And did the same for `build_chunks` one.

usl_pipeline/study_area_uploader/study_area_uploader/chunkers/study_area_chunkers.py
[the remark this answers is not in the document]
waltaskew: Retrieving the chunk size from metadata in the cloud function makes sense to me. Do you have a plan for how to get that into the metastore?

usl_pipeline/study_area_uploader/tests/chunkers/test_study_area_chunkers.py
rsutormin: Done

usl_pipeline/study_area_uploader/tests/chunkers/test_study_area_chunkers.py
```diff
@@ -0,0 +1,255 @@
...
+            soil_classes_polygons=[(bbox_polygon(2, 4, 3, 5), 9)],
+        )
+        assert_check_tar(
+            chunks_dir / "chunk_0_1.tar",
+            2,
+            3,
+            3.0,
+            2.0,
+            [
+                [3.0, 4.0],
```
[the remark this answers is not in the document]
rsutormin: Added padding

usl_pipeline/study_area_uploader/study_area_uploader/chunkers/study_area_chunkers.py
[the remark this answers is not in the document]
rsutormin: I guess we can add some new field called `chunk_size` and update it during the chunking stage?

usl_pipeline/study_area_uploader/study_area_uploader/main.py
```diff
@@ -91,18 +102,19 @@ def export_to_city_cat(
...
-    # Place-holder for the real chunking function.
-    tar_fd = io.BytesIO()
-    with tarfile.open(mode="w", fileobj=tar_fd) as tar:
-        tar.add(
-            str(prepared_inputs.elevation_file_path), arcname=file_names.ELEVATION_TIF
-        )
-    tar_fd.flush()
-    tar_fd.seek(0)
-    yield tar_fd
+    study_area_chunkers.build_and_upload_chunks(
```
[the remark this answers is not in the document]
rsutormin: Got rid of it (and the `export_to_citycat` as well).

usl_pipeline/study_area_uploader/study_area_uploader/chunkers/study_area_chunkers.py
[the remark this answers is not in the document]
waltaskew: yep, sounds good
that does require all chunks being the same size (or else each chunk would attempt to set a different chunk_size for the study area if it's assuming it's own size is the correct size) which lines up with what we'd decided previously.

usl_pipeline/study_area_uploader/study_area_uploader/chunkers/study_area_chunkers.py
```diff
@@ -0,0 +1,180 @@
...
+                band=input_elevation_band,
+                no_data_value=default_nodata_value,
+            )
+        if (
+            chunk_elevation.header.row_count < chunk_size
+            or chunk_elevation.header.col_count < chunk_size
+        ):
+            y_pad = chunk_size - chunk_elevation.header.row_count
+            x_pad = chunk_size - chunk_elevation.header.col_count
+            chunk_elevation.data = numpy.pad(
```
waltaskew, line 134: Is this logic that should be part of `elevation_chunkers.split_geotiff_into_chunks` itself?

rsutormin: I thought about it but then decided that we never read elevation GeoTiff directly from python inside `elevation_chunkers.split_geotiff_into_chunks`, we do `GDAL.translate` to cut chunks, nothing else. On the contrary, in the uploader script I'm reading chunk elevation data anyway to repack the band number and default NODATA. So I'm optimizing the number of read/writes with this one. Let me know if you would want to move the padding deeper into elevation chunker.

waltaskew: sounds good -- might just leave a comment with some thoughts.

waltaskew (approved):
LGTM -- hooray for uploads!

usl_pipeline/study_area_uploader/study_area_uploader/chunkers/study_area_chunkers.py
rsutormin: Done. Also I swapped X and Y axis indices in chunk naming to match it to your schema.

PR #70, opened 2024-06-06
[WRF] Add initial feature transform logic for WPS variables

saskatchuwan:
Following requirements defined in: https://www.notion.so/AtmoML-Model-Inputs-Understanding-WPS-d70ccb1787324b2f9d2aa617413e31b8

* Include `PRES`, `GHT`, `RH`, `TT` to initial list of required WPS vars
* All vars that include `num_metgrid_levels` spatial dim, extract to first level only
* All vars that have percentage unit, convert to decimal
* Apply global normalization


usl_pipeline/cloud_functions/main.py
```diff
@@ -21,7 +21,7 @@
 from usl_lib.storage import cloud_storage
 from usl_lib.storage import file_names
 from usl_lib.storage import metastore
-from usl_lib.shared import wps_variables
+from usl_lib.shared.wps_data import ML_REQUIRED_VARS_REPO, ScalingType, Unit
```
waltaskew, line 24: the style guide sez:
"Use import statements for packages and modules only, not for individual types, classes, or functions."

https://google.github.io/styleguide/pyguide.html#22-imports

as does the zen of Python:

```
$ python3
Python 3.11.8 (main, Feb  7 2024, 21:52:08) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import this
The Zen of Python, by Tim Peters
...
Namespaces are one honking great idea -- let's do more of those!
```

usl_pipeline/cloud_functions/main.py
```diff
@@ -320,21 +320,50 @@ def _read_wps_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
...
-    feature_vals = numpy.squeeze(feature, axis=0)
+    # Drop time axis
+    feature = feature.isel(Time=0)
+
+    # FNL-derived var - extract to only first level
+    if "num_metgrid_levels" in feature.dims:
+        feature = feature.isel(num_metgrid_levels=0)
+
+    feature_values = feature.values
+    wps_var = ML_REQUIRED_VARS_REPO.get(feature.name)
```
waltaskew, line 341: if we're calling `.get` here, does that mean `wps_var` could be `None`? It looks like we don't try to handle that case below.

usl_pipeline/cloud_functions/main.py
```diff
@@ -320,21 +320,50 @@ def _read_wps_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
...
+    return x / 100
+
+
+def _apply_minmax_scaler(x, minx, maxx):
+    # Clip values if they go out of bounds of the specified min and max
+    x[x > maxx] = maxx
+    x[x < minx] = minx
 
-    return feature_vals
+    return (x - minx) / (maxx - minx)
```
waltaskew, line 366: you are officially a data scientist now!

waltaskew (commented):
This looks like a good way to handle the different feature flavors!

usl_pipeline/cloud_functions/main.py
waltaskew: Ah, it does in fact look like MyPy has noticed this as well!

usl_pipeline/cloud_functions/main.py
```diff
@@ -16,12 +16,12 @@
 import xarray
 import rasterio
 
+import usl_lib.shared.wps_data as wps_data
```
waltaskew, line 19: should move this between
from usl_lib.readers import elevation_readers
&
from usl_lib.storage import cloud_storage

to keep the import alphabetized. I'm surprised there isn't a lint for this.

waltaskew (approved):
LGTM

PR #72, opened 2024-06-10
[WRF] Fix the order of netcdf spatial dimensions to always be: (lon, lat, ...)

saskatchuwan:
Variables in WRF/WPS will always be associated with some list of dimensions. The list of dimensions associated with each variable may vary depending on the nature of the variable. Typically, the dimensions could be `Time`, `south_north` (lon), `west_east` (lat), `some-z-dim`, `some-z-dim-1`, etc

Ordering of the dimensions follow a general convention of `Time`, `z`, `lon`, `lat` but may not always be consistent and the z dimension may also vary - see more: https://www.bic.mni.mcgill.ca/users/sean/Docs/netcdf/guide.txn_22.html.

We'll add an initial transform to set the axis order to always be `west_east` (lon), `south_north` (lat), `<other_spatial_dims>` to ensure consistency when processing the feature matrix (especially when extracting values from certain z dimensions, ex: `num_metgrid_levels`)

### DIFF
--- a/usl_pipeline/cloud_functions/main_test.py
+++ b/usl_pipeline/cloud_functions/main_test.py
@@ -236,8 +236,11 @@ def test_build_feature_matrix_wrf(
     # Ensure we attempted to upload a serialized matrix of the netcdf
     mock_feature_blob.upload_from_file.assert_called_once_with(mock.ANY)
     uploaded_array = numpy.load(mock_feature_blob.upload_from_file.call_args[0][0])
-    # Expected array should be all required vars extracted and stacked
-    expected_array = numpy.dstack((netcdf_array1, netcdf_array2))
+    # Expected array should be all required vars extracted, lon/lat axis reordered, and
+    # stacked
+    expected_array = numpy.dstack(
+        (numpy.swapaxes(netcdf_array1, 0, 1), numpy.swapaxes(netcdf_array2, 0, 1))
+    )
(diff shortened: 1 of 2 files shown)


usl_pipeline/cloud_functions/main_test.py
```diff
@@ -165,12 +165,16 @@ def test_build_feature_matrix_wrf(
     time = ncfile.createVariable("time", "f8", ("Time",))
     lat = ncfile.createVariable("lat", "float32", ("west_east",))
     lon = ncfile.createVariable("lon", "float32", ("south_north",))
-    # Create dataset entries for all variables in mock
+    # Create dataset entries for all variables in mock. Order of the lat/lon dims
```
waltaskew, line 168: Why does swapping the lat lons make for easier testing? Is that why we added the `feature = feature.transpose("west_east", "south_north", ...)` call above? I'd generally recommend testing with data that looks as much like the real data as possible.

saskatchuwan: Ack - I agree on using real-ish data during test. My rationale was that it would help separate testing concerns since `process_wps_feature` will perform a series of transforms that are additive (for ex: reordering axis), I wanted to make it easier to isloation the transform logic that we were testing.

I can refacotr to make all the test datasets mimic the real inputs though

waltaskew (approved):
LGTM

PR #69, opened 2024-06-05
Cloud function logic to build feature matrices

rsutormin:

PR #74, opened 2024-06-12
The fix for the wrong CF flood logic updating metadata elevation min

rsutormin:


usl_pipeline/cloud_functions/main.py
```diff
@@ -387,19 +387,20 @@ def _read_elevation_features(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
...
-        rasterio.io.MemoryFile(fd.read())
-    ).data
+    elevation = elevation_readers.read_from_geotiff(rasterio.io.MemoryFile(fd.read()))
 
-    if elevation is None:
+    if elevation.data is None:
         raise ValueError("Elevation file unexpectedly empty.")
 
+    # NODATA cells should be excluded from min/max calculation
+    present_data = elevation.data[elevation.data != elevation.header.nodata_value]
```
waltaskew, line 396: This fixes the min & max problem, but we still should not be passing -9999s in the elevation forward to ML, as I noted over here:
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/69#discussion_r1631591204

Passing -9999s forward in the elevation data is the underlying problem we need to address.

rsutormin: It doesn't matter what you pass for NODATA cells. The values we pass for NODATA cells should not participate in scaling at all. They should be excluded and overwritten by 0 at the end of scaling. Setting 0 values instead of -9999 here is just wrong approach because min might be negative (see our Manhattan case: https://screenshot.googleplex.com/4zJ8q6jikcyL27K).

waltaskew: It matters very much what we pass for nodata! If we send the model a tensor with -9999s, it will have a bad time with them!

I don't think it particularly matters what value we use for no-data, since that information is conveyed by the no-data mask, but if it is outside the 0-1 range we're scaling to, the model is going to have problems.

waltaskew: > They should be excluded and overwritten by 0 at the end of scaling.

Is the assumption that the modelling code will be responsible for reading the elevation data, reading the no-data mask, and zero-ing elevation data using the no-data mask? I'd really like to minimize the amount of work done in the modelling code. We'd also need a way for the modelling code to tell which is the elevation layer and which is the no-data layer, which gets us back to the HDF5 conversation.

rsutormin: Walt, what value do you want me to put instead of -9999 for NODATA cells? If you want me to put 0 in a hope that scaling will be applied to them without taking into account mask layer then I disagree because elevation data includes negative values for Manhattan (https://screenshot.googleplex.com/4zJ8q6jikcyL27K). Just let me know what approach you're proposing here.

waltaskew (approved):

PR #75, opened 2024-06-13
[WRF] Create Cloud Function to build and upload study area chunk for wps files

saskatchuwan:
Create a new cloud function that we'll use to read in uploaded raw WPS files and chunk them into TAR files and upload into `climateiq-study-area-chunks` bucket.

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -138,6 +138,37 @@ def _retry_and_report_errors(
 
 @functions_framework.cloud_event
 @_retry_and_report_errors()
+def build_and_upload_study_area_chunk(
+    cloud_event: functions_framework.CloudEvent,
+) -> None:
+    """Creates a study area chunk and uploads it to GCS.
+
+    This function is triggered when files containing raw geo data for a study area
+    are uploaded to GCS. It will group the files into TAR-files and upload it to a
+    bucket.
+    """
+    file_name = cloud_event.data["name"]
+    bucket_name = cloud_event.data["bucket"]
+
+    if re.search(file_names.WPS_DOMAIN3_NC_REGEX, file_name):
+        storage_client = storage.Client()
+        bucket = storage_client.bucket(bucket_name)
+        file_blob = bucket.blob(file_name)
+
+        # For WRF, we can treat each snapshot output file as its own chunk. If
+        # multiple snapshots must be processed together, we can choose to group
+        # the files together in a TAR before uploading.
+        with file_blob.open("rb") as f:
+            study_area_chunk_bucket = storage_client.bucket(
+                cloud_storage.STUDY_AREA_CHUNKS_BUCKET
+            )
+            study_area_chunk_bucket.blob(file_name).upload_from_file(f)
+
+        # TODO: Write to metastore - create new chunk entry
+
+
+@functions_framework.cloud_event
+@_retry_and_report_errors()
 def write_study_area_metadata(cloud_event: functions_framework.CloudEvent) -> None:
     """Writes metadata for a study area on upload.
 
@@ -511,7 +542,8 @@ def _build_feature_matrix_from_archive(
             if name == file_names.ELEVATION_TIF:
                 return _read_elevation_features(fd)
             # Handle heat model input files (WPS)
-            elif name.startswith("met_em") and name.endswith(".nc"):
+            # TODO: Refactor to also handle un-tarred files
+            elif re.search(file_names.WPS_DOMAIN3_NC_REGEX, name):
                 return _build_wps_feature_matrix(fd)
             # TODO: handle additional archive members.
             else:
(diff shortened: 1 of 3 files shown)


usl_pipeline/usl_lib/usl_lib/storage/file_names.py
```diff
@@ -10,3 +10,5 @@
 CITYCAT_BUILDINGS_TXT = "Buildings.txt"
 CITYCAT_GREEN_AREAS_TXT = "GreenAreas.txt"
 CITYCAT_SPATIAL_GREEN_AREAS_TXT = "Spatial_GreenAreas.txt"
+
+WPS_DOMAIN3_NC_REGEX = "met_em.d03.*.nc$"
```
waltaskew, line 14: I think you'd need to escape those `.`s which are meant to be periods? You can also use r-strings to avoid Python escaping on to of regex escaping.
```python
WPS_DOMAIN3_NC_REGEX = r"met_em\.d03.*\.nc$"
```

usl_pipeline/cloud_functions/main.py
```diff
@@ -136,6 +136,44 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
+    are uploaded to GCS. It will group the files into TAR-files and upload it to a
+    bucket.
+    """
+    file_name = cloud_event.data["name"]
+    bucket_name = cloud_event.data["bucket"]
+
+    if re.search(file_names.WPS_DOMAIN3_NC_REGEX, file_name):
+        storage_client = storage.Client()
+        bucket = storage_client.bucket(bucket_name)
+        file_blob = bucket.blob(str(file_name))
```
waltaskew, line 156: it looks like `file_name` should already be a string? Why the `str` call?

usl_pipeline/cloud_functions/main.py
```diff
@@ -136,6 +136,44 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
+
+        with file_blob.open("rb") as input_file:
+            tar_fd = io.BytesIO()
+            with tarfile.open(mode="w", fileobj=tar_fd) as tar:
+                nc = tarfile.TarInfo(file_name)
+                nc.size = len(input_file)
+                tar.addfile(nc, io.BytesIO(input_file))
```
waltaskew, line 163: You should be able to just pass `input_file` in there. I actually wouldn't expect wrapping it in a BytesIO object to work because the BytesIO constrcutor expects bytes, not a file
https://docs.python.org/3/library/io.html#io.BytesIO

waltaskew (commented):
It looks like this has us storing the chunks twice, once as the raw files, and once with the raw files wrapped in tars. Could we just store it once?

waltaskew:
I think it's totally fine if the feature generation code handles both tars and regular files. We could have code like:

```python
if is_tar(file):
  for member in file:
    if is_elevation(member):
      process_elevation(member)
    elif is_wps(file):
      process_wps(file)
elif is_elevation(file):
  process_elevation(file)
elif is_wps(file):
  process_wps(file)
```
or more generally:
```python
if is_tar(file):
  for member in file:
    process(file)
else:
  process(file)
```
We don't need to make the input be a tar just because that's what we do for flood

saskatchuwan:
> It looks like this has us storing the chunks twice, once as the raw files, and once with the raw files wrapped in tars. Could we just store it once?

that's correct - this is mostly to separate storage concerns (one bucket for all wrf/wps raw files, including those not used by ml model and another that is for ml feature chunks). if we wanted to store only once, the only options i see are:
1. Keep all WPS/WRF file in separate `raw-wrf-sim-data` bucket, and separately install `build_feature_matrix` CF to this bucket
2. Upload and store all WPS/WRF files instead to `cilmateiq-study-area-chunks`, and `build_feature_matrix` CF works as is

WDYT

waltaskew:
I'm fine with either option 1 or 2 -- up to you! It's pretty easy to trigger the same cloud function on multiple buckets, so neither strikes me as being particularly more operationally difficult than the other.

waltaskew:
> Oh - something I thought of too in favor of TARing WPS files is in the future when we build out way to aggregate multiple output files together to make one matrix into one day.

What I meant to get across is that it's easy for us to write code that supports either the file itself or the file inside a tar, so that shouldn't be the thing forcing us into a tar. We should be able to support both pretty easily.

Just don't want 'it needs to be in a tar' as the reason for storing things twice, once as a tar once not.

saskatchuwan:
> > Oh - something I thought of too in favor of TARing WPS files is in the future when we build out way to aggregate multiple output files together to make one matrix into one day.
> 
> What I meant to get across is that it's easy for us to write code that supports either the file itself or the file inside a tar, so that shouldn't be the thing forcing us into a tar. We should be able to support both pretty easily.
> 
> Just don't want 'it needs to be in a tar' as the reason for storing things twice, once as a tar once not.

Ah gotcha - personally, I don't actually prefer alternatives 1 or 2 over the current setup we have for flood that we're using for wrf and primary reason being is we can keep the data processing for all hazard models consistent and predictable this way (easier to onboard new model in the future).

That said, agree with you that we can easily update `build_feature_matrix` to process tar or untarred files and since it's not totally necessary right now for wrf/wps files to be tarred, I can just remove the tarring step here.

PR #76, opened 2024-06-14
Apply processing for vars with `z-dimension0012`

saskatchuwan:
For vars with `z-dimension0012` - each index will correspond to a month (0 for Jan, 11 for Dec). For dimension, we want to extract the value associated with the month of the current WPS output datetime.

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -524,6 +524,9 @@ def _build_wps_feature_matrix(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
     # Ignore type checker error - BytesIO inherits from expected type BufferedIOBase
     # https://shorturl.at/lk4om
     with xarray.open_dataset(fd) as ds:  # type: ignore
+        # Assign Time coordinates so datetime is associated with each data array
+        ds = ds.assign_coords(Time=ds.Times)
+
         features_components = []
         for var_name in wps_data.ML_REQUIRED_VARS_REPO.keys():
             var_config = wps_data.ML_REQUIRED_VARS_REPO[var_name]
@@ -552,9 +555,7 @@ def _process_wps_feature(feature: xarray.DataArray, var_config: dict) -> NDArray
       A new numpy array with transforms applied according to rules
       for each variable defined in: https://shorturl.at/W6nzY
     """
-    # Drop time axis
-    feature = feature.isel(Time=0)
-
+    snapshot_time = feature.coords.get("Time")
     # Ensure order of dimension axes are: (west_east, south_north, <other spatial dims>)
     # to stay consistent with rest of data pipeline
     feature = feature.transpose("west_east", "south_north", ...)
@@ -562,6 +563,18 @@ def _process_wps_feature(feature: xarray.DataArray, var_config: dict) -> NDArray
     # FNL-derived var - extract to only first level
     if "num_metgrid_levels" in feature.dims:
         feature = feature.isel(num_metgrid_levels=0)
+
+    # Monthly climatology var - extract to only the month of the file's datestamp
+    if "z-dimension0012" in feature.dims and snapshot_time:
+        snapshot_datetime = datetime.datetime.strptime(
+            snapshot_time.values[0].decode(), "%Y-%m-%d_%H:%M:%S"
+        )
+        feature = feature.isel({"z-dimension0012": snapshot_datetime.month - 1})
+
+    # At this point, we can remove the Time axis from current feature DataArray
+    # since we no longer need it for processing or in the feature matrix for ML
+    # model. xarray.DataArry.isel() will return new DataArray with axis dropped.
+    feature = feature.isel(Time=0)
 
     feature_values = feature.values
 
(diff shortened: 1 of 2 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -563,6 +564,16 @@ def _process_wps_feature(feature: xarray.DataArray, var_config: dict) -> NDArray
...
         feature = feature.isel(num_metgrid_levels=0)
 
+    # Monthly climatology var - extract to only the month of the file's datestamp
+    if "z-dimension0012" in feature.dims and snapshot_time:
+        snapshot_datetime = datetime.datetime.strptime(
+            snapshot_time.values[0].decode(), "%Y-%m-%d_%H:%M:%S"
+        )
+        feature = feature.isel({"z-dimension0012": snapshot_datetime.month - 1})
+
+    # Drop time axis, no longer needed
```
waltaskew, line 574: Was it ever needed? Does the code still work if `feature = feature.isel(Time=0)` is up in its original location since we're getting the time from `feature.coords` ?

usl_pipeline/cloud_functions/main.py
```diff
@@ -563,6 +564,16 @@ def _process_wps_feature(feature: xarray.DataArray, var_config: dict) -> NDArray
     if "num_metgrid_levels" in feature.dims:
         feature = feature.isel(num_metgrid_levels=0)
 
+    # Monthly climatology var - extract to only the month of the file's datestamp
+    if "z-dimension0012" in feature.dims and snapshot_time:
+        snapshot_datetime = datetime.datetime.strptime(
+            snapshot_time.values[0].decode(), "%Y-%m-%d_%H:%M:%S"
+        )
+        feature = feature.isel({"z-dimension0012": snapshot_datetime.month - 1})
```
waltaskew, line 572: What's the `- 1` in `snapshot_datetime.month - 1` here?

usl_pipeline/cloud_functions/main.py
```diff
@@ -563,6 +564,16 @@ def _process_wps_feature(feature: xarray.DataArray, var_config: dict) -> NDArray
     if "num_metgrid_levels" in feature.dims:
         feature = feature.isel(num_metgrid_levels=0)
 
+    # Monthly climatology var - extract to only the month of the file's datestamp
+    if "z-dimension0012" in feature.dims and snapshot_time:
+        snapshot_datetime = datetime.datetime.strptime(
+            snapshot_time.values[0].decode(), "%Y-%m-%d_%H:%M:%S"
```
waltaskew, line 570: Do we have a consistent name to get this value from? `snapshot_time.values[0]` is going to be grabbing an item at random. Is it always going to be a single value? Is that something we should assert?

saskatchuwan: yeah this is always going to be a single valued array since each wps file corresponds to only one time value

usl_pipeline/cloud_functions/main.py
saskatchuwan: we subtract one here to get the corresponding index. For example if the snapshot month is July (7), then the value from the z-dim we want will be index=6

usl_pipeline/cloud_functions/main.py
saskatchuwan: I moved this towards the bottom once we're finished with transforms that need reference to Time (for exaple, this monthly map extraction). If we keep it at the top, we'll remove the Time dim and the corresponding coords metadata

saskatchuwan: 1. Attach the wps snapshot as a coordinate (so that we can reference it as metadata attached at the DataArray level) - without this step, snapshot time is only availlable at dataset-level
```py
ds = ds.assign_coords(Time=ds.Times)
```

2. Get the time and do something with it during feature engineering if the feature eng step requires ref to snpashot time (for ex: processing z-dim)
```py
snapshot_time = feature.coords.get("Time")
```

3. At some point at the end of feature eng processing, we can drop the time axis (shape=1) since we don't need that extra dimension in the numpy array for ml modeling. this is performed on all feature matrices
```py
feature = feature.isel(Time=0)
```
^ currently, I'm using this because it was easy to use this xarray method to remove a specifc axis using a standard name instead of just an index. But could choose to do this all the way at the end (line 625) if we wanted to. LMK if that helps

waltaskew: The thing I'm specifically confused about is when you say

> we'll remove the Time dim and the corresponding coords metadata

that makes it sounds like `feature = feature.isel(Time=0)` removes the time from the coordinates. Is that true? Can you still access `feature.coords.get("Time")` after doing `feature = feature.isel(Time=0)` ?

saskatchuwan: > that makes it sounds like feature = feature.isel(Time=0) removes the time from the coordinates. Is that true?
Yeah it removes time from the feature dataarray

> Can you still access feature.coords.get("Time") after doing feature = feature.isel(Time=0) ?
No so that's why i moved `feature = feature.isel(Time=0)` from the top since we wanted to do some feature processing that still need reference to the time

waltaskew: Okay, so would that affect other features, then? If we loop over a second feature, would it still be able to access `feature.coords.get("Time")` if the previous loop removed it? Or is `ds = ds.assign_coords(Time=ds.Times)` effectively giving each feature its own copy of the times?

I think this would be helpful to document with some more comments.

saskatchuwan: Ah I see your concern. No, it will not affect other features. `feature = feature.isel(Time=0)` does not mutate the feature DataArray it operates on, it returns a new DataArray. I'll add a comment!

waltaskew (approved):
LGTM

PR #77, opened 2024-06-17
Support for HDF5 file read/write in usl_lib

rsutormin:


usl_pipeline/usl_lib/usl_lib/readers/hdf5_readers.py
```diff
@@ -0,0 +1,29 @@
...
+
+import h5py
+import numpy
+import numpy.typing as npt
+
+
+def read_from_hdf5(
+    input_stream: typing.BinaryIO,
+    group_name: str = "features",
+) -> Dict[str, npt.NDArray[numpy.float32]]:
```
waltaskew, line 12: Not all of the layers are going to be float32 (the masks will be ints)
I'd just leave it as a generic array.

usl_pipeline/usl_lib/usl_lib/readers/hdf5_readers.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Returns:
+        Dictionary with keys corresponding to feature layer names and values containing
+        Numpy 2d arrays of each feature layer.
+    """
+    hf = h5py.File(input_stream, "r")
+    group = hf.get(group_name)
+    layers: Dict[str, npt.NDArray[numpy.float32]] = {}
+    for key, layer in group.items():
+        layers[key] = numpy.array(layer, dtype=numpy.float32)
```
waltaskew, line 27: As above, we won't want to convert the masks to floats. I'd just leave them as-is and assume they've been written with the correct dtype.

usl_pipeline/usl_lib/usl_lib/writers/hdf5_writers.py
```diff
@@ -0,0 +1,26 @@
...
+    Args:
+        layers: Dictionary with keys corresponding to feature layer names and values
+            containing Numpy 2d arrays of each feature layer.
+        output_stream: Output byte stream to write to.
+        group_name: Name of HDF5 group to attach layers to.
+    """
+    hf = h5py.File(output_stream, "w")
+    group = hf.create_group(group_name)
+    for key, layer in layers.items():
+        group.create_dataset(key, data=layer)
```
waltaskew, line 25: Why would we write these as datasets within a group versus have each item in the layers dictionary correspond to a top-level group?

waltaskew (commented):
Do we need this interface? The hdf5 interface already exposes hdf5 files as dictionary-like objects. 

I also am not sure if we should be nesting datasets under groups rather than just working with top-level groups, which essentially treats each hdf5 as a dictionary, rather than a dictionary-of-dictionaries that come from combining groups and datasets.

usl_pipeline/usl_lib/usl_lib/writers/hdf5_writers.py
```diff
@@ -0,0 +1,26 @@
...
+
+import h5py
+import numpy
+import numpy.typing as npt
+
+
+def write_to_hdf5(
+    layers: Dict[str, npt.NDArray[numpy.float32]],
+    output_stream: typing.BinaryIO,
+    group_name: str = "features",
```
waltaskew, line 12: Might not even make this a parameter to prevent folks from being able to create dictionary-of-dictionary structures with hdf5 files containing multiple groups.

usl_pipeline/usl_lib/usl_lib/readers/hdf5_readers.py
```diff
@@ -0,0 +1,29 @@
...
+
+    Args:
+        input_stream: Input byte stream to read from.
+        group_name: Name of HDF5 group where layers are stored.
+
+    Returns:
+        Dictionary with keys corresponding to feature layer names and values containing
+        Numpy 2d arrays of each feature layer.
+    """
+    hf = h5py.File(input_stream, "r")
```
waltaskew, line 23:
```python
with h5py.File(input_stream, "r") as hf
```
would prevent us from having to `hf.close()` ourselves.

usl_pipeline/usl_lib/usl_lib/readers/hdf5_readers.py
rsutormin: Done

usl_pipeline/usl_lib/usl_lib/readers/hdf5_readers.py
rsutormin: I checked, when you return group itself it becomes unconnected from the data after the h5py.File object gets closed. So this conversion to Dict of Numpy is needed for later usage.

usl_pipeline/usl_lib/usl_lib/writers/hdf5_writers.py
rsutormin: Switched to a constant.

usl_pipeline/usl_lib/usl_lib/readers/hdf5_readers.py
rsutormin: Done

rsutormin:
> Do we need this interface? The h5py interface already exposes hdf5 files as dictionary-like objects.

I would prefer this to be a one-liner in the giant CF/main.py file (btw, I would vote for decomposing it at least into flood and wrf parts).

waltaskew (changes requested):
This all LGTM, but I'd like to let Ahmed work on this & get experience with things like managing requirements.txt

PR #81, opened 2024-06-18
Adding elevation header upload as JSON-file in study_area_uploader

rsutormin:
This is the first half of the change (on the uploader side). CF part will go as a separate PR.


waltaskew (approved):
LGTM

PR #85, opened 2024-06-21
[Flood] Fix for the elevation min/max metadata bug in case of empty study area

rsutormin:


usl_pipeline/cloud_functions/main.py
```diff
@@ -690,6 +690,26 @@ def _apply_minmax_scaler(x, minx, maxx):
...
+    """Calculates metadata for the elevation data."""
+    elevation_data = elevation.data
+
+    if elevation_data is None:
+        raise ValueError("Elevation file unexpectedly empty.")
+
+    # NODATA cells should be excluded from min/max calculation
+    present_data = elevation_data[elevation_data != elevation.header.nodata_value]
+
+    if len(present_data) == 0:
```
waltaskew, line 703: `len` returns the length of the first axis, so you want to check `size`
```python
if present_data.size == 0
```
if you want to check the number of elements in the array. They're equivalent here because present_data should be one dimensional, but I think `.size` here would be more idiomatic.  `len(arr)` is really just `arr.shape[0]` which I think is confusing enough that it's worth generally avoiding.

waltaskew (approved):
LGTM -- just a small thing about `len` in numpy

usl_pipeline/cloud_functions/main.py
rsutormin: Done

PR #86, opened 2024-06-21
[WRF] Create new study area metastore entry when raw WPS file is uploaded

saskatchuwan:
When raw WPS files are first uploaded into `climateiq-atmospheric-simulation-input` GCS bucket, in-scope WPS file will be uploaded as a study area chunk. When this occurs, we will additionally write to metastore to create a new entry for this study area.

CRS value will be derived from a cartopy projection that utilize's wps file's map_projection value (lambert, meridian, etc). We will use the source_crs to populate the metastore entry

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -16,6 +16,8 @@ from google.cloud import storage
 import functions_framework
 import numpy
 from numpy.typing import NDArray
+import netCDF4
+import wrf
 import xarray
 import rasterio
 from shapely import geometry
@@ -149,30 +151,60 @@ def _retry_and_report_errors(
 def build_and_upload_study_area_chunk(
     cloud_event: functions_framework.CloudEvent,
 ) -> None:
-    """Creates a study area chunk and uploads it to GCS.
+    """Creates a study area chunk, uploads to bucket, and writes to metastore.
 
     This function is triggered when files containing raw geo data for a study area
-    are uploaded to GCS. It will group the files into TAR-files and upload it to a
-    bucket.
+    are uploaded to GCS. It will load the file into study area chunks bucket and
+    write a new study area to metastore.
     """
-    file_name = cloud_event.data["name"]
+    file_name = pathlib.PurePosixPath(cloud_event.data["name"])
     bucket_name = cloud_event.data["bucket"]
 
-    if re.search(file_names.WPS_DOMAIN3_NC_REGEX, file_name):
+    # For AtmoML, only 3rd domain (500m) output files will be used
+    if re.search(file_names.WPS_DOMAIN3_NC_REGEX, file_name.name):
         storage_client = storage.Client()
+        db = firestore.Client()
+
         bucket = storage_client.bucket(bucket_name)
-        file_blob = bucket.blob(file_name)
+        file_blob = bucket.blob(str(file_name))
 
         # For WRF, we can treat each snapshot output file as its own chunk. If
         # multiple snapshots must be processed together, we can choose to group
         # the files together in a TAR before uploading.
-        with file_blob.open("rb") as f:
+        with file_blob.open("rb") as fd:
             study_area_chunk_bucket = storage_client.bucket(
                 cloud_storage.STUDY_AREA_CHUNKS_BUCKET
             )
-            study_area_chunk_bucket.blob(file_name).upload_from_file(f)
+            study_area_chunk_bucket.blob(str(file_name)).upload_from_file(fd)
 
-        # TODO: Write to metastore - create new chunk entry
+            # File names should be in the form <study_area_name>/<file_name>
+            study_area_name = file_name.parent.name
+
+            nc_bytes = fd.read()
+            with netCDF4.Dataset("in_memory.nc", memory=nc_bytes) as ds:
+                study_area = metastore.StudyArea(
+                    name=study_area_name,
+                    # Grid dimension will always be 200x200 for all cities
+                    col_count=200,
+                    row_count=200,
+                    x_ll_corner=ds.getncattr("corner_lons")[0],
+                    y_ll_corner=ds.getncattr("corner_lats")[0],
+                    # https://www2.mmm.ucar.edu/wrf/users/namelist_best_prac_wps.html#dx_dy
+                    cell_size=int(ds.getncattr("DX")),
+                    # https://www2.mmm.ucar.edu/wrf/users/namelist_best_prac_wps.html#map_proj
+                    crs=_get_crs_from_wps(ds),
+                )
+                study_area.set(db)
+
+
+def _get_crs_from_wps(nc_dataset: netCDF4.Dataset) -> str:
+    # The var used here doesn't matter - as long as it is a var that has coordinate
+    # attributes
+    wps_var = wrf.getvar(wrfin=nc_dataset, varname="XLAT_M")
+    cart_proj = wrf.get_cartopy(wps_var)
+    source_crs = cart_proj.source_crs.to_json_dict()["conversion"]["method"]["id"]
+    crs = source_crs["authority"] + ":" + str(source_crs["code"])
+    return crs
 
 
 @functions_framework.cloud_event
--- a/usl_pipeline/cloud_functions/requirements.txt
+++ b/usl_pipeline/cloud_functions/requirements.txt
@@ -50,6 +50,7 @@ python-dateutil==2.9.0.post0
 pytz==2024.1
 rasterio==1.3.9
 requests==2.31.0
+rioxarray==0.15.7
 rsa==4.9
 shapely==2.0.3
 six==1.16.0
--- a/usl_pipeline/cloud_functions/main_test.py
+++ b/usl_pipeline/cloud_functions/main_test.py
@@ -615,21 +615,32 @@ def test_build_feature_matrix_errors(
 
 @mock.patch.object(main.firestore, "Client", autospec=True)
 @mock.patch.object(main.storage, "Client", autospec=True)
-def test_build_and_upload_study_area_chunk(mock_storage_client, mock_firestore_client):
+@mock.patch.object(main, "_get_crs_from_wps", autospec=True)
+def test_build_and_upload_study_area_chunk(
+    mock_get_crs_from_wps, mock_storage_client, mock_firestore_client
+):
+    # Get some random data to place in a netcdf file
+    lat_vals = [40, 45]
+    long_vals = [-79, -80]
+    cell_size = 500
+    source_crs = "EPSG:32618"
+
     # Create an in-memory netcdf file and grab its bytes
     ncfile = netCDF4.Dataset("met_em.d03_test.nc", mode="w", format="NETCDF4", memory=1)
-    ncfile.createDimension("Time", 1)
-    ncfile.createVariable("time", "f8", ("Time",))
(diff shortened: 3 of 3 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -143,30 +143,53 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
 
-        # TODO: Write to metastore - create new chunk entry
+            # File names should be in the form <study_area_name>/<file_name>
+            study_area_name = file_name.parent.name
+
+            # Write metastore entry for study area if it does not exist
+            study_area_exists = (
+                metastore.StudyArea.get_ref(db, str(study_area_name)).get().exists
+            )
+            if not study_area_exists:
```
waltaskew, line 179: The check for existence here would still be vulnerable to race conditions. So instead, I just wouldn't bother.

It's fine to call `set` multiple times.
You can call `set({blah}, merge=True)` to perform an update if you're worried about removing keys set by other processes. e.g.
```
doc.set({'a': 1}, merge=True)
doc.set({'b': 2}, merge=True)
```
would result in `{'a': 1, 'b': 2}` (without `merge=True` it'd just be `{'b': 2}`)

But I think a regular set without the merge is still fine here.

usl_pipeline/cloud_functions/main.py
```diff
@@ -143,30 +143,53 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
+                with xarray.open_dataset(fd) as ds:
+                    x_ll_meters = ds.data_vars["XLONG_M"].values[0, 0, 0] * 111
+                    y_ll_meters = ds.data_vars["XLAT_M"].values[0, 0, 0] * 111
+
+                    study_area = metastore.StudyArea(
+                        name=study_area_name,
+                        col_count=200,
+                        row_count=200,
```
waltaskew, line 187: Is WRF always 200x200? Or is it just 200 for our particular New York study area?

usl_pipeline/cloud_functions/main.py
```diff
@@ -143,30 +143,53 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
+            # File names should be in the form <study_area_name>/<file_name>
+            study_area_name = file_name.parent.name
+
+            # Write metastore entry for study area if it does not exist
+            study_area_exists = (
+                metastore.StudyArea.get_ref(db, str(study_area_name)).get().exists
+            )
+            if not study_area_exists:
+                with xarray.open_dataset(fd) as ds:
+                    x_ll_meters = ds.data_vars["XLONG_M"].values[0, 0, 0] * 111
```
waltaskew, line 181: Why times 111 here? A comment would be helpful here.

usl_pipeline/cloud_functions/main.py
```diff
@@ -143,30 +143,53 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
+                    y_ll_meters = ds.data_vars["XLAT_M"].values[0, 0, 0] * 111
+
+                    study_area = metastore.StudyArea(
+                        name=study_area_name,
+                        col_count=200,
+                        row_count=200,
+                        x_ll_corner=int(x_ll_meters),
+                        y_ll_corner=int(y_ll_meters),
+                        cell_size=250000,
+                    )
```
waltaskew, line 191: we'll also need to set a CRS so we know how to interpret the x & y ll_corner coordinates

usl_pipeline/cloud_functions/main.py
MadhaviJain90: Its always going to be 200x200 for all cities

usl_pipeline/cloud_functions/main.py
MadhaviJain90: This is still WIP. The rule of thumb for converting meters to degrees is to use this conversion: 111kilometer = 1 degree. But this is only a rough estimate and changes with latitude. I wouldn't use this tbh. It'll be helpful to know how x_ll and y_ll are mapped out in FloodML

usl_pipeline/cloud_functions/main.py
MadhaviJain90: I had a question about cell_size. Is it supposed to be cell area? Because thats what the number 250000 means here. Its a 500m by 500m cell length.

waltaskew: Good thing you asked! It should be 500, not 500 x 500

usl_pipeline/cloud_functions/main.py
waltaskew: it's part of the input data for flood -- we have a .tiff with x_ll & y_ll & CRS as an input to define the raster

waltaskew: the x_ll & y_ll are ultimately what we'll send to ClimaSens to build the dashboards, so it will need to be accurate!

waltaskew: for Manhattan we have
```
cell_size: 2
col_count: 5835
crs: "EPSG:32618"
row_count: 10945
x_ll_corner: 580459.5669010617
y_ll_corner: 4504014.133973204
```

MadhaviJain90: This is really helpful. cell_size in that case should be 500 (I would remove the meter squared comment -> its misleading)
col_count and row_count will be 200 for all cities
I need to do some digging to be sure about the crs and the x_ll and y_ll

saskatchuwan: Thanks @waltaskew @MadhaviJain90 - updated. I still have an open about CRS tho - @MadhaviJain90 how do we find out what system we use for WRF?

waltaskew: At a glance, looks like the CRS can be found using this:
https://wrf-python.readthedocs.io/en/latest/user_api/generated/wrf.get_cartopy.html#wrf-get-cartopy

x_ll_corner and y_ll_corner are the coordinates of the lower left corner of the lower left cell. They don't need to be converted into meters by multiplying by 111. Keeping them as degrees is fine.

waltaskew: Also, want to confirm (0, 0) is the lower left corner and not (0, 200) -- can double check the lat lon values to see where the mins are

usl_pipeline/cloud_functions/main.py
```diff
@@ -143,30 +143,49 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
+                    name=study_area_name,
+                    # Grid dimension will always be 200x200 for all cities
+                    col_count=200,
+                    row_count=200,
+                    x_ll_corner=ds.attrs["corner_lons"][0],
+                    y_ll_corner=ds.attrs["corner_lats"][0],
+                    # https://www2.mmm.ucar.edu/wrf/users/namelist_best_prac_wps.html#dx_dy
+                    cell_size=int(ds.attrs["DX"]),
+                    # https://www2.mmm.ucar.edu/wrf/users/namelist_best_prac_wps.html#map_proj
+                    crs=ds.attrs["MAP_PROJ"],
```
waltaskew, line 186: That looks like the projection, not the CRS.

This
https://wrf-python.readthedocs.io/en/latest/user_api/generated/wrf.get_cartopy.html#wrf-get-cartopy
returns a `cartopy.crs.Projection` which looks compelling!

saskatchuwan: gotcha - but what value are we trying to extract from that projection?

waltaskew: The CRS itself -- it should be a value like WGS84 or EPSG:32618 or NAD83 or such.

From your link, it looks like MAP_PROJ is one of 'lambert', 'polar', 'mercator', and 'lat-lon'

If you can poke around and look at the values, hopefully it will become apparent if it's a CRS or not.

usl_pipeline/cloud_functions/main.py
waltaskew: So at the end of the day, we're going to use the CRS to pull the predictions into geopandas here:
https://github.com/UrbanSystemsLab/climateiq-frontend/blob/2f54adf0615ee0a9d170b27550ede602972361c9/cloud_functions/climateiq_data_export_cf/main.py#L260
and then export them to ClimaSens.
Can you confirm that geopandas is happy taking in a value like 'lambert' and such? Or if we need to supply a value like 'EPSG:9802' instead? (Could just hop into the unit tests for the above code and swap 'lambert' in as the CRS and see what happens to test this quickly.)

saskatchuwan: No it look slike geopands will expect an authority string: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.html

I've added the code to get the source crs from the wps files though. pleas take a looK!

MadhaviJain90 (approved):
LGTM :)

usl_pipeline/cloud_functions/requirements.txt
```diff
@@ -56,4 +57,5 @@ tzdata==2024.1
 urllib3==2.2.1
 watchdog==4.0.0
 Werkzeug==3.0.1
+wrf-python==1.3.4.1
```
waltaskew, line 60: Noted elsewhere, but we'll need to wait to merge until we know how this impacts cloud functions, as this file is used to deployments right now.

usl_pipeline/cloud_functions/main.py
```diff
@@ -149,30 +151,59 @@ def wrapper(cloud_event: functions_framework.CloudEvent) -> None:
...
+    are uploaded to GCS. It will load the file into study area chunks bucket and
+    write a new study area to metastore.
     """
-    file_name = cloud_event.data["name"]
+    file_name = pathlib.PurePosixPath(cloud_event.data["name"])
     bucket_name = cloud_event.data["bucket"]
 
-    if re.search(file_names.WPS_DOMAIN3_NC_REGEX, file_name):
+    # For AtmoML, only 3rd domain (500m) output files will be used
+    if re.search(file_names.WPS_DOMAIN3_NC_REGEX, file_name.name):
```
waltaskew, line 164: Will we have multiple uploads of multiple uploads of different files for the same area, thus causing this to trigger multiple times and recreate the same study area?

That's fine, but might be worth documenting.

Is there an easy way to only run this function once? (Check if the study area is already in the metatore, or only trigger on file 0 or such?)

waltaskew:
code LGTM, but I don't want to approve & merge until we have WRF requirements.txt cloud function stuff sorted.

usl_pipeline/cloud_functions/main.py
saskatchuwan: Yeah that's right - I originally had a check in here to check if the study area existed before writing but you had mentioned that the multiple writes shouldn't be too big a deal. personally, i'd prefer if we'd only write once

usl_pipeline/cloud_functions/requirements.txt
saskatchuwan: Ack

usl_pipeline/cloud_functions/main.py
waltaskew: it really is not a big deal to set multiple times -- would only suggest changing things if it's a really easy change
I think it's fine to ride with what you've got down already

PR #87, opened 2024-06-24
[WRF] Refactor build_feature_matrix to process un-tarred WRF input files

saskatchuwan:
Since we are treating one WPS file as one study area chunk, there is no need at this point to package this file into an archive for processing. Instead, we will load one WPS file (un-tarred) as a chunk and trigger `build_feature_matrix()` to process the netcdf file.

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -297,9 +297,9 @@ def delete_flood_scenario_metadata(cloud_event: functions_framework.CloudEvent) 
     )
 )
 def build_feature_matrix(cloud_event: functions_framework.CloudEvent) -> None:
-    """Builds a feature matrix when an archive of geo files is uploaded.
+    """Builds a feature matrix when an a set of geo files is uploaded.
 
-    This function is triggered when archive files containing geo data are uploaded to
+    This function is triggered when files containing geo data are uploaded to
     GCS. It produces a feature matrix for the geo data and writes that feature matrix to
     another GCS bucket for eventual use in model training and prediction.
     """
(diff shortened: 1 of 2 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -313,15 +313,23 @@ def build_feature_matrix(cloud_event: functions_framework.CloudEvent) -> None:
...
-            raise ValueError(f"Empty archive found in {chunk_blob}")
+    with chunk_blob.open("rb") as chunk:
+        # Flood (CityCat)
+        if chunk_name.endswith(".tar"):
+            feature_matrix, metadata = _build_feature_matrix_from_archive(chunk)
+
+            if feature_matrix is None:
+                raise ValueError(f"Empty archive found in {chunk_blob}")
+        # Heat (WRF)
+        elif re.search(file_names.WPS_DOMAIN3_NC_REGEX, chunk_name):
```
waltaskew, line 329: It'd also be nice to handle WPS files inside tars. If we can, I'd try to structure this as something like
```python
def _build_feature_matrix(bucket_name, chunk_name, output_bucket):
  if chunk_name.endswith(".tar"):
    tensors = []
    for member in get_tar_members(chunk):
      tensors.append(get_features(chunk))
    return numpy.dstack(tensors)
  else:
    return get_features(chunk)

def get_features(chunk):
  if name == file_names.ELEVATION_TIF:
    return = _read_elevation_features(chunk)
  elif name == file_names.BOUNDARIES_TXT:
    return _read_polygons_from_byte_stream(chunk)
```

usl_pipeline/cloud_functions/main.py
```diff
@@ -574,10 +582,6 @@ def _build_feature_matrix_from_archive(
                 green_areas = _read_polygons_from_byte_stream(fd)
             elif name == file_names.SOIL_CLASSES_TXT:
                 soil_classes = _read_polygons_from_byte_stream(fd)
-            # Handle heat model input files (WPS)
-            # TODO: Refactor to also handle un-tarred files
-            elif re.search(file_names.WPS_DOMAIN3_NC_REGEX, name):
```
waltaskew, line 579: Can we keep this in and support both files in and out of the tar?

usl_pipeline/cloud_functions/main.py
```diff
@@ -603,9 +607,10 @@ def _build_feature_matrix_from_archive(
 
 
 def _build_wps_feature_matrix(fd: IO[bytes]) -> Tuple[NDArray, FeatureMetadata]:
+    fd_bytes = io.BytesIO(fd.read())
     # Ignore type checker error - BytesIO inherits from expected type BufferedIOBase
     # https://shorturl.at/lk4om
-    with xarray.open_dataset(fd) as ds:  # type: ignore
+    with xarray.open_dataset(fd_bytes) as ds:  # type: ignore
```
waltaskew, line 613: Why pass `fd.read()` rather than `fd` here? My hope would be that `open_dataset` can stream the data rather than read it all in at once.

waltaskew (approved):

usl_pipeline/cloud_functions/main.py
saskatchuwan: sure thing - added back in

usl_pipeline/cloud_functions/main.py
saskatchuwan: yeah looks like we can do it if we specify the right engine. updated. tested locally and this should work!

usl_pipeline/cloud_functions/main.py
saskatchuwan: Ack - let me see if we can refactor this a bit more. I'll do it in a separate branch

PR #88, opened 2024-06-24
Fix for the bug in flood chunk CF for the case of empty polygon data

rsutormin:


usl_pipeline/cloud_functions/main.py
```diff
@@ -582,8 +582,22 @@ def _build_feature_matrix_from_archive(
             else:
                 logging.warning(f"Unexpected member name: {name}")
 
-    if any((elevation, buildings, green_areas, soil_classes)):
-        if not all((elevation, buildings, green_areas, soil_classes)):
+    if any(
```
waltaskew, line 585: this might be clearer written out as
```python
files_present = [file is not None for file in (elevation, buildings, green_areas, soil_classes)]
if any(files_present):
  if not all(files_present):
```

waltaskew (approved):
LGTM

usl_pipeline/cloud_functions/main.py
rsutormin: Yep, good idea!

PR #89, opened 2024-06-25
[Flood] First part of feature rescaling (storing chunk info in study area metadata)

rsutormin:
Property "chunk_size" is also going to be used in label chunker instead of hardcoded value.

### DIFF
--- a/usl_pipeline/usl_lib/usl_lib/storage/metastore.py
+++ b/usl_pipeline/usl_lib/usl_lib/storage/metastore.py
@@ -1,5 +1,6 @@
 import dataclasses
 import enum
+import math
 from typing import Iterable, Optional
 import urllib.parse
 
@@ -142,12 +143,10 @@ class StudyArea:
             return
 
         # Calculate number of chunks over each of X and Y axes:
-        x_count = (study_area.col_count + chunk_size - 1) // chunk_size
-        y_count = (study_area.row_count + chunk_size - 1) // chunk_size
+        x_count = math.ceil(study_area.col_count / chunk_size)
+        y_count = math.ceil(study_area.row_count / chunk_size)
 
-        study_area_ref = db.collection(STUDY_AREAS).document(study_area_name)
-        db.transaction().update(
-            study_area_ref,
+        db.collection(STUDY_AREAS).document(study_area_name).update(
             {
                 "chunk_size": chunk_size,
                 "chunk_x_count": x_count,
--- a/usl_pipeline/usl_lib/usl_lib/storage/metastore.py
+++ b/usl_pipeline/usl_lib/usl_lib/storage/metastore.py
@@ -42,6 +42,9 @@ class StudyArea:
     crs: Optional[str] = None
     elevation_min: Optional[float] = None
     elevation_max: Optional[float] = None
+    chunk_size: Optional[int] = None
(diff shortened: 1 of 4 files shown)


usl_pipeline/usl_lib/usl_lib/storage/metastore.py
```diff
@@ -117,6 +120,41 @@ def update_min_max_elevation(
...
+                    + f" {study_area.chunk_size} != {chunk_size}"
+                )
+            return
+
+        # Calculate number of chunks over each of X and Y axes:
+        x_count = (study_area.col_count + chunk_size - 1) // chunk_size
+        y_count = (study_area.row_count + chunk_size - 1) // chunk_size
+
+        study_area_ref = db.collection(STUDY_AREAS).document(study_area_name)
+        db.transaction().update(
```
waltaskew, line 149: It doesn't look like we need a transaction here since we're just performing a single operation. I was using a transaction in `update_min_max_elevation` because I need a transactional read + write.

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
```diff
@@ -117,6 +120,41 @@ def update_min_max_elevation(
...
+            study_area_ref,
+            {
+                "chunk_size": chunk_size,
+                "chunk_x_count": x_count,
+                "chunk_y_count": y_count,
+            },
```
waltaskew, line 155: `ref.set({...}, merge=True)`
will give up update semantics (only setting these three fields, but leaving any other fields alone) without a transaction.

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
```diff
@@ -117,6 +120,41 @@ def update_min_max_elevation(
...
+        if study_area.chunk_size is not None:
+            if study_area.chunk_size != chunk_size:
+                raise ValueError(
+                    "Unexpected variation detected for chunk size:"
+                    + f" {study_area.chunk_size} != {chunk_size}"
+                )
+            return
+
+        # Calculate number of chunks over each of X and Y axes:
+        x_count = (study_area.col_count + chunk_size - 1) // chunk_size
```
waltaskew, line 145: I think the count would be
```
x_count = math.ceil(col_count / chunk_size)
y_count = math.ceil(row_count / chunk_size)
```

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
rsutormin: Yep, SG.

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
rsutormin: I decided to use `ref.update` instead since `set` gives you an advantage of creating the doc if it didn't exist before (which I don't need).

usl_pipeline/usl_lib/usl_lib/storage/metastore.py
rsutormin: Yep, this looks more intuitive.

waltaskew (approved):
LGTM -- hooray for metastorage!

PR #91, opened 2024-06-25
Mark Train / Test Split for Study Area

waltaskew:
- For a study area, indicate which 20% of its chunks should be considered the test set and which 80% should be considered the training set.

@tatianawu @prashantkul 
Here's what I was thinking for the test / train split in the metastore. The study area has a list of `{'x_index': x, 'y_index': y}` pairs for the train & test set. So you can look them up, then select the chunks with those indices within the study area you're interested in working with. Does this look all right to y'all?

### DIFF
--- a/usl_pipeline/cloud_functions/main_test.py
+++ b/usl_pipeline/cloud_functions/main_test.py
@@ -165,9 +165,27 @@ def test_build_feature_matrix_flood(mock_storage_client, mock_firestore_client, 
         mock_firestore_client().collection().document(),
         {"elevation_min": 2, "elevation_max": 6},
     )
-    mock_firestore_client().collection().document().update.assert_called_once_with(
-        {"chunk_size": 2, "chunk_x_count": 5, "chunk_y_count": 10}
-    ),
+
+    # Ensure we called update with the correct dictionary.
+    update_args = (
+        mock_firestore_client().collection().document().update.call_args.args[0]
+    )
+
+    assert update_args.keys() == {
+        "chunk_size",
+        "chunk_x_count",
+        "chunk_y_count",
+        "train_indices",
+        "test_indices",
+    }
+    assert update_args["chunk_size"] == 2
+    assert update_args["chunk_x_count"] == 5
+    assert update_args["chunk_y_count"] == 10
+
+    # 40 / 50 chunks for training
+    assert len(update_args["train_indices"]) == 40
+    # 10 / 50 chunks for training
+    assert len(update_args["test_indices"]) == 10
 
 
 def test_build_feature_matrix_from_archive_empty_polygons():
--- a/usl_pipeline/usl_lib/tests/integration/test_metastore_integration.py
+++ b/usl_pipeline/usl_lib/tests/integration/test_metastore_integration.py
@@ -218,3 +218,42 @@ def test_get_simulation_bad_name_raises_error(firestore_db):
         metastore.Simulation.get(
             firestore_db, "missing-study-area", "missing-config/name.txt"
         )
+
+
+def test_update_chunk_info(firestore_db):
+    metastore.StudyArea(
+        name="study-area",
+        col_count=10,
+        row_count=4,
+        x_ll_corner=3,
+        y_ll_corner=4,
+        cell_size=5,
+        crs="crs",
+    ).create(firestore_db)
+
+    metastore.StudyArea.update_chunk_info(firestore_db, "study-area", 2, seed=2)
+    assert metastore.StudyArea.get(firestore_db, "study-area") == metastore.StudyArea(
+        name="study-area",
+        col_count=10,
+        row_count=4,
+        x_ll_corner=3,
+        y_ll_corner=4,
+        cell_size=5,
+        crs="crs",
+        chunk_x_count=5,  # 10 / 2
+        chunk_y_count=2,  # 4 / 2
+        chunks_size=2,
+        #  8 / 10 chunks for training.
+        train_indices=[
+            {"x_index": 0, "y_index": 0},
+            {"x_index": 0, "y_index": 1},
+            {"x_index": 1, "y_index": 0},
+            {"x_index": 1, "y_index": 1},
+            {"x_index": 2, "y_index": 0},
+            {"x_index": 3, "y_index": 0},
+            {"x_index": 3, "y_index": 1},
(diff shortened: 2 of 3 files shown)


tatianawu:
It looks like we're doing the split based on (x, y) indices for each chunk. Does that mean that, across simulations, the same geographic regions are used for training and testing (i.e., each chunk is always used for either training or testing, and we'll never use it for the other)? If so, that could pose an issue.

prashantkul:
I agree. Test data needs to be random enough as training.



*Prashant Kulkarni*

Customer Engineer, Security & Compliance Specialist

Google Cloud

waltaskew:
> It looks like we're doing the split based on (x, y) indices for each chunk. Does that mean that, across simulations, the same geographic regions are used for training and testing (i.e., each chunk is always used for either training or testing, and we'll never use it for the other)? If so, that could pose an issue.

This makes each 1000x1000 chunk either a training chunk or a test chunk.
What should we be doing instead?

> I agree. Test data needs to be random enough as training.

What do you mean by that?

waltaskew:
@prashantkul @tatianawu Can y'all describe what you want for the test train split here? I thought marking 80% of the chunks as 'train' and 20% of the chunks as 'test' is what y'all had requested. Lemme know how you want this to work, and I'll code it up!

tatianawu:
> @prashantkul @tatianawu Can y'all describe what you want for the test train split here? I thought marking 80% of the chunks as 'train' and 20% of the chunks as 'test' is what y'all had requested. Lemme know how you want this to work, and I'll code it up!

We want an 80/20 split on the **examples**, not the chunks. The clarification I was looking for in my original comment was whether these two are synonymous, because I'm concerned they aren't. If we're indexing on (x, y) chunks, does that mean the same geographical chunk is used for either training or testing, across all examples? For example, does (0, 0) represent a well-defined geographical region? And will it always be used for training?

If so, I'm wondering if we can do the split on the labels instead, since I think these more closely resemble the conventional "example"?

waltaskew:
Ah, I see. 

> > @prashantkul @tatianawu Can y'all describe what you want for the test train split here? I thought marking 80% of the chunks as 'train' and 20% of the chunks as 'test' is what y'all had requested. Lemme know how you want this to work, and I'll code it up!
> 
> We want an 80/20 split on the **examples**, not the chunks. The clarification I was looking for in my original comment was whether these two are synonymous, because I'm concerned they aren't. If we're indexing on (x, y) chunks, does that mean the same geographical chunk is used for either training or testing, across all examples? For example, does (0, 0) represent a well-defined geographical region? And will it always be used for training?
> 
> If so, I'm wondering if we can do the split on the labels instead, since I think these more closely resemble the conventional "example"?

Ah, I see. Yes, at present one chunk is a 1000x1000 area and is associated with a single example. The labels describe the same 1000x1000 chunks.

I agree it makes more sense logically to describe the labels rather than the chunks. It'll come out to the same thing, but would be a bit less confusing!

tatianawu:
> > > @prashantkul @tatianawu Can y'all describe what you want for the test train split here? I thought marking 80% of the chunks as 'train' and 20% of the chunks as 'test' is what y'all had requested. Lemme know how you want this to work, and I'll code it up!
> > 
> > 
> > We want an 80/20 split on the **examples**, not the chunks. The clarification I was looking for in my original comment was whether these two are synonymous, because I'm concerned they aren't. If we're indexing on (x, y) chunks, does that mean the same geographical chunk is used for either training or testing, across all examples? For example, does (0, 0) represent a well-defined geographical region? And will it always be used for training?
> > If so, I'm wondering if we can do the split on the labels instead, since I think these more closely resemble the conventional "example"?
> 
> Ah, I see. Yes, at present one chunk is a 1000x1000 area and is associated with a single example. The labels describe the same 1000x1000 chunks.
> 
> I agree it makes more sense logically to describe the labels rather than the chunks. It'll come out to the same thing, but would be a bit less confusing!

Sorry, can you explain how it's the same thing? I think there's some confusion because our data isn't in well-defined examples.

Say we have 2 different simulations (S1 and S2). Looking at a single chunk C1, that is a geographical area with the same geospatial features. However, rainfall looks different between S1 and S2, which means that we have 2 different examples, correct?

Example 1: Uses C1 and S1 rainfall
Example 2: Uses C1 and S2 rainfall

We have 2 different examples that happen to use the same geospatial chunk/features. Am I understanding correctly that the current train/test splitting approach wouldn't make this distinction? It should be possible to have example 1 used for training and example 2 used for testing.

waltaskew:
> > > > @prashantkul @tatianawu Can y'all describe what you want for the test train split here? I thought marking 80% of the chunks as 'train' and 20% of the chunks as 'test' is what y'all had requested. Lemme know how you want this to work, and I'll code it up!
> > > 
> > > 
> > > We want an 80/20 split on the **examples**, not the chunks. The clarification I was looking for in my original comment was whether these two are synonymous, because I'm concerned they aren't. If we're indexing on (x, y) chunks, does that mean the same geographical chunk is used for either training or testing, across all examples? For example, does (0, 0) represent a well-defined geographical region? And will it always be used for training?
> > > If so, I'm wondering if we can do the split on the labels instead, since I think these more closely resemble the conventional "example"?
> > 
> > 
> > Ah, I see. Yes, at present one chunk is a 1000x1000 area and is associated with a single example. The labels describe the same 1000x1000 chunks.
> > I agree it makes more sense logically to describe the labels rather than the chunks. It'll come out to the same thing, but would be a bit less confusing!
> 
> Sorry, can you explain how it's the same thing? I think there's some confusion because our data isn't in well-defined examples.
> 
> Say we have 2 different simulations (S1 and S2). Looking at a single chunk C1, that is a geographical area with the same geospatial features. However, rainfall looks different between S1 and S2, which means that we have 2 different examples, correct?
> 
> Example 1: Uses C1 and S1 rainfall Example 2: Uses C1 and S2 rainfall
> 
> We have 2 different examples that happen to use the same geospatial chunk/features. Am I understanding correctly that the current train/test splitting approach wouldn't make this distinction? It should be possible to have example 1 used for training and example 2 used for testing.

You're right! Yes, this wouldn't capture the distinction between one geography being associated with multiple examples for differing configurations.

waltaskew:
Closing in favor of https://github.com/UrbanSystemsLab/climateiq-cnn/pull/102 which actually does this correctly!

PR #92, opened 2024-06-26
[Flood] Second part of feature rescaling (CF logic)

rsutormin:


usl_pipeline/cloud_functions/main_test.py
```diff
@@ -380,10 +384,14 @@ def test_build_feature_matrix_wrf(mock_storage_client, mock_firestore_client, _)
                 {
                     # For heat, we process each WPS netcdf file as a chunk (un-tar'ed)
                     "archive_path": "gs://bucket/study_area/met_em.d03_test.nc",
-                    "feature_matrix_path": (
+                    "unscaled_feature_matrix_path": (
```
rsutormin, line 387: Surprisingly my change in the study area chunk metadata code is involved into WRF too! Walt, I guess we need to discuss how to decouple Flood and WRF here...

usl_pipeline/cloud_functions/main.py
```diff
@@ -862,3 +869,80 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+        with feature_blob.open("rb") as in_fd, output_blob.open("wb") as out_fd:
+            feature_raster_transformers.rescale_feature_blob(in_fd, study_area, out_fd)
+    # Once all chunks are processed we can delete unscaled ones:
+    for feature_blob in feature_blobs:
+        feature_blob.delete()
+
+
+@functions_framework.cloud_event
+@_retry_and_report_errors()
+def rescale_feature_matrices(cloud_event: functions_framework.CloudEvent) -> None:
```
waltaskew, line 932: What were you envisioning would trigger this function?
If the unscaled outputs are written to the features bucket, it will trigger the `build_feature_matrix` function again. Triggers run for all uploads to anywhere in the bucket. You can't have a trigger on certain paths. 
For the collapsing the labels, I rely on the fact that step 1 has .rsl files as input and step 2 has .npy files as input, so there's one cloud function, but it dispatches based on the file type
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/ed43aa9375d33d33d65d71a7c47d172075e72092/usl_pipeline/cloud_functions/main.py#L372

I think we'd do similar here -- have one function that first checks to see if the file path is for a scaled or unscaled file and dispatch based on that

waltaskew:
So I think the general flow I'd shoot for here is:
```python
def build_feature_matrix(cloud_event):
  blob = bucket.blob(cloud_event.data["name"])
  if is_flood_file(blob.name):
    build_flood_matrix(blob)
  elif is_atmo_file(blob.name):
    build_atmo_matrix(blob)

def build_flood_matrix(blob):
  if is_raw_feature_input_name(blob.name):
    write_feature_matrix(blob)
  elif is_unscaled_feature_name(blob.name):
    write_scaled_inputs(blob)
  elif is_scaled_feature_name(blob.name):
    return
```
where you could write a file like 'unscaled_features.npy' and the trigger could notice the file is named 'unscaled_features.npy' and try to scale it if all the expected numpy arrays are present.

Writing the scaled features array will of course trigger the function again, so we'd need to be sure to ignore that write!

usl_pipeline/cloud_functions/main.py
rsutormin: Walt, I think you're mixing up buckets. `build_feature_matrix` CF is triggered when new file is stored to `study-area-chunks` bucket. This new CF function (`rescale_feature_matrices`) is going to be attached to `study-area-feature-chunks` bucket which doesn't currently have any attached triggers (as far as I know). Please correct me if I'm wrong.

usl_pipeline/cloud_functions/main_test.py
rsutormin: I ended up with using this check to separate flood from WRF: https://github.com/UrbanSystemsLab/climateiq-cnn/pull/92/files#diff-fed8bb97ec2754b34cfbc4861b38234014e6cf77913263164691eeb39c8be94eR331

usl_pipeline/cloud_functions/main.py
waltaskew: Ah! Okay, that's why I was asking -- I wasn't sure if you were planning to trigger this function off of study-area-chunks or study-area-feature-chunks writes.
Yes, you're correct that that we don't have any triggers for feature-chunks yet.

waltaskew: We will want to avoid applying flood logic (like this scaling) to heat files written to the study-area-feature-chunks bucket as well.

rsutormin:
> So I think the general flow I'd shoot for here is:
> 
> ```python
> def build_feature_matrix(cloud_event):
>   blob = bucket.blob(cloud_event.data["name"])
>   if is_flood_file(blob.name):
>     build_flood_matrix(blob)
>   elif is_atmo_file(blob.name):
>     build_atmo_matrix(blob)
> 
> def build_flood_matrix(blob):
>   if is_raw_feature_input_name(blob.name):
>     write_feature_matrix(blob)
>   elif is_unscaled_feature_name(blob.name):
>     scale_features_if_all_are_present(blob)
>   elif is_scaled_feature_name(blob.name):
>     return
> ```
> 
> where you could write a file like 'unscaled_features.npy' and the trigger could notice the file is named 'unscaled_features.npy' and try to scale it if all the expected numpy arrays are present.
> 
> Writing the scaled features array will of course trigger the function again, so we'd need to be sure to ignore that write!

I have this logic checking `chunk_*` prefix in the new CF (see here: https://github.com/UrbanSystemsLab/climateiq-cnn/pull/92/files#diff-fed8bb97ec2754b34cfbc4861b38234014e6cf77913263164691eeb39c8be94eR903). And also I just added new test that makes sure this logic is not applied to `scaled_*` files (see https://github.com/UrbanSystemsLab/climateiq-cnn/pull/92/files#diff-b91f014c9f83a783e5e5dea25e4d37c17686a840d34690c2be676b4e8340b136R1495-R1501)

usl_pipeline/cloud_functions/main.py
rsutormin: I think I didn't change current WRF logic by my PR. So I guess let's separate refactoring effort decoupling the logic between Flood and WRF into separate PR? I don't want to change WRF logic in this PR.

usl_pipeline/usl_lib/usl_lib/transformers/feature_raster_transformers.py
```diff
@@ -127,3 +129,55 @@ def transform_to_feature_raster_layers(
...
+        study_area_metadata: Study Area metadata containing min/max elevation values
+            that are used for scaling.
+
+    Returns:
+        Scaled feature matrix.
+    """
+    shape = unscaled_feature_matrix.shape
+    feature_layers = [
+        layer.reshape(shape[0], shape[1])
+        for layer in numpy.dsplit(unscaled_feature_matrix, shape[2])
```
waltaskew, line 171: We don't need to dsplit the arrays then dstack them back together. We can select along the third dimension directly. So we can do
```python
elevation_data = unscaled_feature_matrix[:, :, 0]
presence_mask = unscaled_feature_matrix[:, :, 1]
elevation_data[:] = elevation_data[presence_max == 1] = 
        elevation_data[presence_mask == 1] - elevation_min
    ) / (elevation_max - elevation_min)
```
which would mutate `unscaled_feature_matrix` in place.

rsutormin: Done

usl_pipeline/cloud_functions/main.py
```diff
@@ -862,3 +882,80 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+    for feature_blob in feature_blobs:
+        logging.info(f"[Feature Rescaler] Rescaling feature matrix {feature_blob.path}")
+        output_blob = feature_bucket.blob(
+            f"{feature_parent_path}/scaled_{feature_blob.name}"
+        )
+        with feature_blob.open("rb") as in_fd, output_blob.open("wb") as out_fd:
+            feature_raster_transformers.rescale_feature_blob(in_fd, study_area, out_fd)
+    # Once all chunks are processed we can delete unscaled ones:
+    for feature_blob in feature_blobs:
+        feature_blob.delete()
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/cloud_functions/main.py
```diff
@@ -808,17 +812,33 @@ def _write_flood_chunk_metastore_entry(
...
         # Remove any errors from previous failed retries which have now succeeded.
         error=firestore.DELETE_FIELD,
-    ).merge(db, study_area_name)
+    )
+    if support_scaled_features:
+        study_area_chunk.unscaled_feature_matrix_path = (
+            f"gs://{feature_blob.bucket.name}/{feature_blob.name}"
+        )
+        feature_blob_file_name = pathlib.PurePosixPath(feature_blob.name).name
+        study_area_chunk.feature_matrix_path = (
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/cloud_functions/main.py
rsutormin: Yep, `needs_scaling` field was added.

usl_pipeline/cloud_functions/main.py
```diff
@@ -903,3 +900,98 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+    """
+    study_area_name, chunk_name = _parse_chunk_path(blob_path)
+    # Let's look up chunk metadata and check if scaling is needed
+    db = firestore.Client()
+    chunk_metadata = metastore.StudyAreaSpatialChunk.get_if_exists(
+        db, study_area_name, chunk_name
+    )
+    if chunk_metadata is None:
+        logging.info(f"Chunk metadata was not found for {study_area_name}/{chunk_name}")
+        return
```
waltaskew, line 928: should this be an error condition? should we raise something here?

usl_pipeline/cloud_functions/main.py
```diff
@@ -903,3 +900,98 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+        logging.info(f"Chunk {study_area_name}/{chunk_name} doesn't need scaling")
+        return
+
+    study_area = metastore.StudyArea.get(db, study_area_name)
+    if study_area.chunk_x_count is None or study_area.chunk_y_count is None:
+        logging.info(
+            f"[Feature Rescaler] Study area {study_area_name} has no chunk counts in "
+            + "metadata"
+        )
+        return
```
waltaskew, line 940: Should we raise something here?

usl_pipeline/cloud_functions/main.py
```diff
@@ -903,3 +900,98 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+        output_blob = feature_bucket.blob(
+            f"{study_area_name}/scaled_{blob_chunk_name}.npy"
+        )
+        with feature_blob.open("rb") as in_fd, output_blob.open("wb") as out_fd:
+            feature_raster_transformers.rescale_feature_blob(in_fd, study_area, out_fd)
+
+    logging.info("[Feature Rescaler] Updating feature matrix path in chunk metadata...")
+    for feature_blob in feature_blobs:
+        _, blob_chunk_name = _parse_chunk_path(feature_blob.name)
+        scaled_feature_matrix_path = (
```
waltaskew, line 966: Should we stick this logic in the loop above? Or did you want to explicitly avoid writing any metadata until all chunks are scaled?

usl_pipeline/cloud_functions/main.py
```diff
@@ -903,3 +900,98 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+        return
+
+    # We're at the final step when all unscaled feature matrices are ready
+    for feature_blob in feature_blobs:
+        logging.info(f"[Feature Rescaler] Rescaling feature matrix {feature_blob.name}")
+        _, blob_chunk_name = _parse_chunk_path(feature_blob.name)
+        output_blob = feature_bucket.blob(
+            f"{study_area_name}/scaled_{blob_chunk_name}.npy"
+        )
+        with feature_blob.open("rb") as in_fd, output_blob.open("wb") as out_fd:
```
waltaskew, line 960: By its nature, this will be a run-at-least-once rather than run-exactly-once operation. We could try to scale the same files twice. We should add similar exception handling as I have here:
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_pipeline/cloud_functions/main.py#L503
in case we're running twice and the other invocation has started deleting chunks.

usl_pipeline/cloud_functions/main.py
rsutormin: Yeah, I'm thinking about retry logic that will try to fix some error if it's related to scaling. If we delete some part of original unscaled files retry logic wouldn't be possible to proceed.

usl_pipeline/cloud_functions/main.py
rsutormin: No, this may happen for scaled files for instance (these files have names completely unrelated to chunk_id in metadata. So this is expected case. I just wanted to see in logs that everything is going as expected.

usl_pipeline/usl_lib/usl_lib/transformers/feature_raster_transformers.py
```diff
@@ -127,3 +129,54 @@ def transform_to_feature_raster_layers(
...
+    scaled_feature_matrix = rescale_feature_matrix(
+        unscaled_feature_matrix, study_area_metadata
+    )
+    numpy.save(feature_matrix_output_fd, scaled_feature_matrix)
+
+
+def rescale_feature_matrix(
+    feature_matrix: npt.NDArray,
+    study_area_metadata: metastore.StudyArea,
+) -> npt.NDArray:
```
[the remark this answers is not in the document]
rsutormin: Done

usl_pipeline/cloud_functions/main.py
rsutormin: Yeah, I think you're right, this is not expected situation (initially I thought some chunks that are processed too quickly before study area metadata is processed might not see that metadata yet).

usl_pipeline/cloud_functions/main.py
rsutormin: Actually, I think it's much safer to trigger scaling processing separately for each chunk. See my recent refactoring related to this.

usl_pipeline/cloud_functions/main.py
rsutormin: I ended up with quite a refactoring. Now the idea is we generate `trigger` files for each feature matrix file. Each trigger-file triggers separate CF processing for corresponding feature matrix.

usl_pipeline/cloud_functions/main.py
```diff
@@ -903,3 +903,104 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+                + " in metadata"
+            )
+
+        feature_blobs = list(
+            feature_bucket.list_blobs(prefix=f"{study_area_name}/chunk_")
+        )
+
+        chunk_count = study_area.chunk_x_count * study_area.chunk_y_count
+        if chunk_count != len(feature_blobs):
+            raise ValueError(
```
waltaskew, line 953: Is this an error condition, or are we just waiting for the rest of the files to be written?

usl_pipeline/cloud_functions/main.py
waltaskew: Cloud functions by their nature are run-at-least-once rather than run-exactly-once, so we could still end up with two invocations trying to scale the same one file, so we could potentially still see a deletion while trying to read here. Certainly less likely with your refactoring, though!

usl_pipeline/cloud_functions/main.py
waltaskew: sounds good -- might add a comment explaining this

usl_pipeline/cloud_functions/main.py
```diff
@@ -903,3 +903,104 @@ def _write_as_npy(blob: storage.Blob, array: NDArray) -> None:
...
+        # Deleting trigger file
+        feature_bucket.blob(blob_path).delete()
+
+
+@functions_framework.cloud_event
+@_retry_and_report_errors()
+def rescale_feature_matrices(cloud_event: functions_framework.CloudEvent) -> None:
+    """Starts rescaling process once all unscaled feature matrix files are uploaded.
+
+    Output files with scaled feature matrices will be placed to the same folder where
```
waltaskew, line 993: Let's describe the scale trigger files as well. I'd generally suggest being aggressive with the amount of comments you write in here.

waltaskew (approved):
hooray for parallel scaling!

I'll also want to use the same trigger for recognizing that all feature files have been written to handle breaking them into separate train & validation sets.

usl_pipeline/cloud_functions/main.py
rsutormin: Oh, sorry, you're right. It was originally done as logging and then I lost track a bit. Added a comment here to keep the context later.

usl_pipeline/cloud_functions/main.py
rsutormin: I added the try-except-not-found check that you proposed above.

usl_pipeline/cloud_functions/main.py
rsutormin: Done

PR #93, opened 2024-06-26
Add early stopping support to FloodML

tatianawu:
Added optional early stopping to the FloodML model.

Also fixed a bug in the FloodModelParams dataclass. The default optimizer is being used across all instances of the dataclass that don't specify a different value. Because this is currently a keras.Optimizer object, that same object is getting passed to possibly multiple models, which will throw an error. Changed this to a string, which is also a valid optimizer value. As with all other model parameters, we leave it to the user to supply reasonable values.

### DIFF
--- a/usl_models/tests/model_test.py
+++ b/usl_models/tests/model_test.py
@@ -1,5 +1,6 @@
 """Tests for Flood model."""
 
+import dataclasses
 import pytest
 
 import tensorflow as tf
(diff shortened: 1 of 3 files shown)


usl_models/tests/model_test.py
```diff
@@ -183,3 +184,38 @@ def test_bad_labels():
...
+def test_early_stopping():
+    """Tests early stopping during model training.
+
+    Since inputs are random, the model will not continually improve (there's no real
+    training). Setting a reasonably large number of epochs makes it likely that the
+    validation loss stops improving before training is complete. Enabling early stopping
+    with a patience of 1 epoch means that the model should stop training as soon as the
+    validaton loss goes up.
+
+    Note that this test is nondeterministic and possibly flaky. But with a large enough
```
waltaskew, line 198: If we set a random seed such that we're consistently using the same "random" inputs every time, I think that would address possible flakiness?
https://www.tensorflow.org/api_docs/python/tf/random/set_seed

tatianawu: You're right! Thanks for the suggestion, added a seed.

waltaskew (approved):
LGTM

PR #95, opened 2024-06-28
create atmo_ml Folder

SallyElHajjar:


tatianawu:
It looks like you're mixing up two different GitHub accounts. The one you're logged into on browser is different from the one that's pushing the code.

SallyElHajjar:
t seems when I had authenticate my git (log time ago) I used my gmail. Is that possible to proceed by doing that?

tatianawu:
Please use the same account. I don't think there's any reason to have multiple GitHub accounts. You can add multiple emails to the same account, and configure your settings to push to a repo using a specific email address. You probably cloned the repo and authenticated using the wrong account in your terminal.

@waltaskew Would you recommend ignoring this for now and submitting the PR? Or should we fix the multiple accounts first?

waltaskew:
> Please use the same account. I don't think there's any reason to have multiple GitHub accounts. You can add multiple emails to the same account, and configure your settings to push to a repo using a specific email address. You probably cloned the repo and authenticated using the wrong account in your terminal.

I'd also recommend using one GitHub account and using [github's email routing configuration](https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#choosing-where-your-organizations-email-notifications-are-sent) so that notifications for USL repos go to your USL email address rather than creating a USL GitHub account.

[SallyElHajjar](https://github.com/SallyElHajjar) is the account currently configured as a collaborator for this repo. I can change it to a different account if you like, but this is currently the correct GitHub account to use for submitting code to this repo.

> @waltaskew Would you recommend ignoring this for now and submitting the PR? Or should we fix the multiple accounts first?

I wouldn't worry about it too much for this PR. I think the main thing to do is
- figure out which account you want to use
- for the email associated with that account, do
```
git config --global user. email "you@example.com"
```
to use it going forward with git
- let me know which GitHub account is the correct one to configure as a collaborator for this repo. Currently it is the [SallyElHajjar](https://github.com/SallyElHajjar) account on this pull request.

PR #96, opened 2024-06-28
Store Chunk Metadata for Flood & WRF

waltaskew:
- Disentangle the WRF & flood feature logic a bit.
- For flood, write metadata for spatial chunks which include their indices within the study area.
- For flood, write metadata for temporal chunks which include their time.

PR #97, opened 2024-07-01
[WRF] Create new cloud function to build WRF label matrix

saskatchuwan:
* Process only domain3 wrfout files
* Use wrf-python to derive variables of interest: RH2, T2, Wind speed and direction
* Update SimulationLabelChunk entry in metastore for each label matrix built


usl_pipeline/cloud_functions/main.py
```diff
@@ -291,6 +293,83 @@ def delete_flood_scenario_metadata(cloud_event: functions_framework.CloudEvent)
...
+    """Updates the metastore with new information for the given wrf chunks.
+
+    Args:
+      chunk_blob: The GCS blob containing the raw chunk.
+      label_blob: The GCS blob containing the label tensor for the chunk.
+    """
+    db = firestore.Client()
+    study_area_name = _parse_chunk_path(chunk_blob.name)[0]
+
+    metastore.SimulationLabelChunk(
```
saskatchuwan, line 365: @waltaskew Do we want a WRF-specific simulation label chunk object here?

waltaskew: ah, yes, I think we'll want to mirror
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/96
in that we're dividing not dividing these up spatially as we do for flood

saskatchuwan: Ack - I can throw that in this PR!

usl_pipeline/cloud_functions/requirements.txt
```diff
@@ -35,7 +35,7 @@ itsdangerous==2.1.2
 Jinja2==3.1.3
 MarkupSafe==2.1.5
 netCDF4==1.6.5
-numpy==1.26.4
+numpy>=1.25.2
```
saskatchuwan, line 38: @waltaskew Do you know a better way to resolve numpy install issue that's failing the build?

I changed this line because of subprocess error when installing `wrf-python` - this lib will also install numpy as a dep but expects it to be a a different version. pip install runs fine when I install the requirements in a local env but something in the github action is not happy..

waltaskew: we should set the right numpy version in here, then. We want these to be exact `==` versions to avoid unexpected version changes in deployment.

The process for updating this file should be
```python
pip install <all the things>
pip freeze > requirements.txt
```
so you'll export whatever versions `pip` resolved the various libraries to

waltaskew: failing how? When I pip install wrf-python and the other things in the requirements file, pip selects `numpy==1.25.2`

saskatchuwan: Tried setting it to 1.15.2 too but:
https://github.com/UrbanSystemsLab/climateiq-cnn/actions/runs/9764950697/job/26954503839?pr=97

MadhaviJain90 (approved):
Looks good!!

usl_pipeline/cloud_functions/requirements.txt
saskatchuwan: This should be fixed once https://github.com/UrbanSystemsLab/climateiq-cnn/pull/108 is merged

waltaskew (commented):
This all looks lovely! Just the same comment again about being reluctant to approve before we know it's safe to merge the requirments.txt file with wrf-python in it.

usl_pipeline/cloud_functions/main.py
```diff
@@ -292,6 +294,88 @@ def delete_flood_scenario_metadata(cloud_event: functions_framework.CloudEvent)
...
+    storage_client = storage.Client()
+    bucket = storage_client.bucket(bucket_name)
+    chunk_blob = bucket.blob(chunk_name)
+
+    with chunk_blob.open("rb") as chunk:
+        # Heat (WRF) - treat one WRF output file as one chunk
+        if re.search(file_names.WRF_DOMAIN3_NC_REGEX, chunk_name):
+            label_matrix, metadata = _build_wrf_label_matrix(chunk)
+
+            label_file_name = chunk_name + ".npy"
```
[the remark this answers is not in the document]
saskatchuwan: Yeah I had to change it specifically bc the wrf out files do not have have a file extension: `wrfout_d01_2023-03-31_06_00_00`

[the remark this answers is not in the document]
saskatchuwan: @MadhaviJain90 This will always be true right? the wrfout files will not have the .nc file extension?

saskatchuwan:
@waltaskew Just double checking that we'll want to impl `is_in_test_set` for the wrf temporal chunks too?

waltaskew:
> @waltaskew Just double checking that we'll want to impl `is_in_test_set` for the wrf temporal chunks too?

Yes, we will. We should chat with Tatiana & Sally about this a bit.

For labels describing spatial chunks, we randomly pick 20% of them as reserved for testing and 80% for training.

For your WRF labels describing temporal chunks, we might want to do something a little different. For instance, we might be concerned about how contiguous the test set is (e.g. the test set could be composed of five day pieces rather than just random days which may not be contiguous.)

usl_pipeline/cloud_functions/main.py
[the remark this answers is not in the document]
waltaskew: could have the code try to detect it there is a stem or not, but if we end up with a file like `blah.nc.npy` that's not a big deal

waltaskew (approved):

usl_pipeline/cloud_functions/main.py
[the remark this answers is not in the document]
saskatchuwan: Cool updated to handle both cases

PR #99, opened 2024-07-01
Add logic to save and load model checkpoints

tatianawu:


usl_models/tests/model_test.py
```diff
@@ -218,3 +219,35 @@ def test_early_stopping():
...
+        batch_size,
+        height=height,
+        width=width,
+        storm_duration=storm_duration,
+        include_labels=True,
+    )
+    fake_data = flood_model.FloodModelData(storm_duration=storm_duration, **fake_input)
+    model.train([fake_data])
+
+    model_path = "/tmp/test_model.keras"
```
waltaskew, line 243: we'll want to use a tempfile to make sure this gets deleted, and also for comparability with our windows friends who don't have a /tmp folder

https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

```python
with tempfile.NamedTemporaryFile() as tmp:
  model.save_model(tmp.name)
  ...
```

usl_models/tests/model_test.py
```diff
@@ -218,3 +219,35 @@ def test_early_stopping():
...
+    model.save_model(model_path, overwrite=True)
+
+    new_model = flood_model.FloodModel(params, spatial_dims=(height, width))
+    new_model.load_model(model_path)
+
+    old_weights = model._model.get_weights()
+    new_weights = new_model._model.get_weights()
+    assert len(old_weights) == len(new_weights)
+    for i, weights in enumerate(old_weights):
```
waltaskew, line 252: looks like we could zip these together and do
```python
for old, new in zip(old_weights, new_weights):
  np.testing.assert_array_equal(old, new_)
```
to avoid working with a loop variable

usl_models/usl_models/flood_ml/model.py
```diff
@@ -177,6 +178,27 @@ def train(
 
         return model_history
 
+    def load_model(self, filepath: str):
```
waltaskew, line 181: can we add an annotation for the return type? Is it `-> 'FloodModel:'` ?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -177,6 +178,27 @@ def train(
...
+        """Loads weights from an existing file.
+
+        Args:
+            filepath: Path to the weights file to load into the current model.
+        """
+        self._model.load_weights(filepath)
+        logging.info("Loaded model weights from %s", filepath)
+
+    def save_model(self, filepath: str, overwrite: bool = False):
```
waltaskew, line 190: Need to add an explicit `-> None:` to allow MyPy to pay attention to the function.

usl_models/tests/model_test.py
tatianawu: Done.

usl_models/tests/model_test.py
tatianawu: Good callout, fixed!

usl_models/usl_models/flood_ml/model.py
tatianawu: Oops, the annotations didn't get pushed. This should be None.

usl_models/usl_models/flood_ml/model.py
tatianawu: Done.

waltaskew (approved):
LGTM

PR #102, opened 2024-07-03
Mark Label Chunks Test or Train

waltaskew:
- Add an 'in_test_set' field to our label chunks.
- Mark each chunk as in the test set or not.

### DIFF
--- a/usl_pipeline/cloud_functions/main_test.py
+++ b/usl_pipeline/cloud_functions/main_test.py
@@ -1241,7 +1241,7 @@ def test_collapse_city_cat_output_chunks(mock_firestore_client):
                     ),
                     "x_index": 0,
                     "y_index": 1,
-                    "in_validation_set": False,
+                    "in_test_set": False,
                 }
             ),
         ]
--- a/usl_pipeline/cloud_functions/main_test.py
+++ b/usl_pipeline/cloud_functions/main_test.py
@@ -1132,12 +1132,25 @@ def test_write_city_cat_output_chunks(mock_firestore_client):
 def test_collapse_city_cat_output_chunks(mock_firestore_client):
     """Ensures we write the correct chunks to GCS."""
     # Have firestore return a three-timestep configuration.
-    mock_firestore_client().collection().document().get().to_dict.return_value = {
-        "gcs_uri": "gs://config-bucket/config/file.txt",
-        "as_vector_gcs_uri": "gs://label-bucket/config/file.npy",
-        "parent_config_name": "config",
-        "rainfall_duration": 3,
-    }
+    mock_firestore_client().collection().document().get().to_dict.side_effect = [
+        # The first call to get is for a FloodScenarioConfig.
+        {
+            "gcs_uri": "gs://config-bucket/config/file.txt",
+            "as_vector_gcs_uri": "gs://label-bucket/config/file.npy",
+            "parent_config_name": "config",
+            "rainfall_duration": 3,
+        },
+        # The second call is for a StudyArea.
+        {
+            "col_count": 2,
+            "row_count": 5,
+            "x_ll_corner": 0.0,
+            "y_ll_corner": 0.0,
+            "cell_size": 1.0,
+            "chunk_x_count": 2,
+            "chunk_y_count": 5,
+        },
+    ]
 
     # Create a blob for the file triggering the cloud function run.
     blob_path = pathlib.PurePosixPath(
@@ -1228,6 +1241,7 @@ def test_collapse_city_cat_output_chunks(mock_firestore_client):
                     ),
                     "x_index": 0,
                     "y_index": 1,
+                    "in_validation_set": False,
                 }
             ),
         ]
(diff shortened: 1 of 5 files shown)


usl_pipeline/usl_lib/tests/storage/test_metastore.py
```diff
@@ -211,3 +211,77 @@ def test_flood_scenario_config_delete():
...
+        cell_size=3.0,
+        crs="over there",
+        chunk_size=1,
+        chunk_x_count=2,
+        chunk_y_count=5,
+    )
+
+    # Calculate whether all chunks are in the validation set.
+    chunks_in_validation_set = [
+        metastore.SimulationLabelChunk.is_in_validation_set(study_area, "config", 0, 0),
```
rsutormin, line 233: Walt, can you also show in test code that several consequent calls to the same parameters return the same result (even though there is random generator involved). I know there is a seed initiated by hasher. But still it's good to double check that everything behaves as expected.

rsutormin (approved):

usl_pipeline/usl_lib/tests/storage/test_metastore.py
waltaskew: Sure, I can do that, although we're essentially doing that because if not for a seed, the tests would be very flaky! :)

waltaskew:
@prashantkul @tatianawu can y'all confirm if this matches the test & validate behaviour y'all are expecting? We'll need to merge this and re-run data pipelines to get these changes out.

prashantkul:
Maybe I didn't understand correctly - are you creating a test/train set or
a validation/train set? They're different and ideally we want all 3 of
them. Training set needs to be randomly shuffled.

waltaskew:
> Maybe I didn't understand correctly - are you creating a test/train set or a validation/train set? They're different and ideally we want all 3 of them. Training set needs to be randomly shuffled.

Sorry, train & validate. When generating the labels, we mark them as either part of train or validate in the data pipeline.
The test & train will need to happen later in the dataset generation.

tatianawu:
> > Maybe I didn't understand correctly - are you creating a test/train set or a validation/train set? They're different and ideally we want all 3 of them. Training set needs to be randomly shuffled.
> 
> Sorry, train & validate. When generating the labels, we mark them as either part of train or validate set in the data pipeline. The test & train will need to happen later in the dataset generation.

Can you explain what you mean by the train/test split needs to happen later?

We need a train/test split. We don't technically need a separate validation set; we can choose to randomly split off part of each batch for validation. I actually think it would be fine to just switch the current validation set you're creating to being the test set.

prashantkul:
I think we want a Train/Test first. Validation can be split off from the
training set. Looks like that was the intention of the PR based on its
title.

prashantkul:
yes, I agree.

waltaskew:
Sorry, I may have my terminology mixed up.
What I'm doing here is marking 20% of labels as 'in_validation_set'
Should I have called it 'in_test_set' instead?

I thought y'all wanted the data pipeline to mark 20% of the labels as something like that?

prashantkul:
ah, it's a terminology issue. Yes it should have been "in_test_set"



*Prashant Kulkarni*

Customer Engineer, Security & Compliance Specialist

Google Cloud

waltaskew:
Okay! Sorry for the confusion! The branch name is `simulation-test-train-indicators` which proves I chose the word 'test' and 'validate' at random.

I've updated the flag to 'in_test_set'

usl_pipeline/cloud_functions/main_test.py
```diff
@@ -1228,6 +1241,7 @@ def test_collapse_city_cat_output_chunks(mock_firestore_client):
                     ),
                     "x_index": 0,
                     "y_index": 1,
+                    "in_validation_set": False,
```
tatianawu, line 1244: Your tests still use `in_validation_set` in some places.

waltaskew: yep, I've now fixed this and the build is happy

usl_pipeline/usl_lib/tests/storage/test_metastore.py
```diff
@@ -211,3 +211,99 @@ def test_flood_scenario_config_delete():
...
+        x_ll_corner=1.0,
+        y_ll_corner=2.0,
+        cell_size=3.0,
+        crs="over there",
+        chunk_size=1,
+        chunk_x_count=2,
+        chunk_y_count=5,
+    )
+
+    # Calculate whether all chunks are in the validation set.
```
tatianawu, line 231: Nit: Some of the comments still refer to this as the validation set, which might cause some confusion.

tatianawu (approved):

usl_pipeline/usl_lib/tests/storage/test_metastore.py
waltaskew: sorry, removed those too

PR #103, opened 2024-07-03
Fix for the bug in chunk metadata logic related to absent  field

rsutormin:


waltaskew (approved):
LGTM -- might add an integration test or refactor in terms of `.get()` later once we've got the bug fix deployed

PR #104, opened 2024-07-03
Fix for the bug in flood feature rescaler CF related to massive triggering

rsutormin:


usl_pipeline/cloud_functions/main.py
```diff
@@ -965,8 +965,14 @@ def _start_feature_rescaling_if_ready(feature_bucket: storage.Bucket, blob_path:
                 chunk_count,
                 len(feature_blobs),
             )
+            return
```
waltaskew, line 968: ah!

waltaskew (approved):
LGTM

PR #108, opened 2024-07-08
[WRF] Install wrf-python

saskatchuwan:
`wrf-python` package owners only provide macos build distribution (https://pypi.org/project/wrf-python/#files). In order for wrf-python to build properly in linux env (in this case, github workflow), we will pre-generate a linux wheel for this package and update requirements.txt to install from this wheel if in linux environment, otherwise, install from PyPi.

Additionally, a couple of callouts - `wrf-python` dependencies did not play well with current versions of `h5py` and `pandas`. Versions for both packages were updated to earlier versions instead.

TODO: Update CF terraform to copy wheels (https://github.com/UrbanSystemsLab/climateiq-terraform/blob/main/modules/data_pipeline/main.tf#L15)


usl_pipeline/cloud_functions/requirements.txt
```diff
@@ -1,16 +1,25 @@
...
 charset-normalizer==3.3.2
 click==8.1.7
 click-plugins==1.1.1
 cligj==0.7.2
 cloudevents==1.10.1
+contourpy==1.2.1
+cycler==0.12.1
 deprecation==2.1.0
 fiona==1.9.6
+flake8==7.1.0
```
waltaskew, line 19: it looks like dev tools like flake8 & MyPy are in here. Let's just keep this to the actual production requirements.

.github/workflows/python-app.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         mypy usl_pipeline/study_area_uploader
 
   cloud_functions:
-    name: Cloud Funcations CI
+    name: Cloud Functions CI
```
waltaskew, line 90: This name is used in github config. You'll notice the CI for this isn't actually running because the name has changed and the config for the repo is apparently referring to the mispelled name.

It's currently

> Cloud Funcations CI Expected — Waiting for status to be reported

indefinitely. I suspect this PR isn't actually working and we'd need to pull in C & Fortran compilers to build the wheel.

saskatchuwan: Reverted the name and seems to run fine. The previous name change triggers a job but sets the old one to pending: https://github.com/UrbanSystemsLab/climateiq-cnn/actions/runs/9846444330/job/27184227122

lmk what you think!

usl_pipeline/cloud_functions/requirements.txt
saskatchuwan: Done!

waltaskew (commented):
If you deploy a cloud function with this requirements.txt, what happens? Does the deployment fail, or is it able to build the wrf wheel?

We'll need to wait to merge this until we have it working with the cloud function environment.

.github/workflows/python-app.yml
waltaskew: Does look like it builds! The image must already have the compilers & such we need.

waltaskew: we could fix the 'Funcations' name in a follow up PR and adjust the GitHub settings to match.

saskatchuwan:
> If you deploy a cloud function with this requirements.txt, what happens? Does the deployment fail, or is it able to build the wrf wheel?
> 
> We'll need to wait to merge this until we have it working with the cloud function environment.

Can we deploy a CF separately from the main codebase here? Can you point me to where we'd want to update teh CF environment? I can try to take a look

waltaskew (approved):
Wonderful -- thanks for all the documentation!

PR #111, opened 2024-07-08
Cleanup of old chunk and feature files before execution of chunker logic in uploader

rsutormin:


usl_pipeline/study_area_uploader/study_area_uploader/main.py
```diff
@@ -76,21 +98,38 @@ def main() -> None:
...
+    bucket: storage.Bucket, prefix: str, override: bool
+) -> None:
+    """Checks if storage bucket folder has files, deletes if allowed or stops if not."""
+    blobs = [blob for blob in bucket.list_blobs(prefix=f"{prefix}")]
+    if len(blobs) == 0:
+        # No files, nothing to check or delete
+        return
+    if not override:
+        # Deletion is not allowed, stop the execution
+        raise ValueError(
```
waltaskew, line 121: I think it'd be better to print a message and exit cleanly in this case rather than raise an exception. The stack trace will make it look like something broke to the user.

waltaskew: You can use [sys.exit](https://docs.python.org/3/library/sys.html#sys.exit) to return non-zero from the script (although I think it's bad form to call sys.exit outside of `main`)

rsutormin: Yep, I refactored that function to print error to logs and return False. In main I do `sys.exit` every time this function returns False.

waltaskew (approved):

PR #112, opened 2024-07-09
Metastore Interactions for Training & Prediction

waltaskew:
Add functions
- get_temporal_feature_metadata
- get_spatial_feature_chunk_metadata
- get_spatial_feature_and_label_chunk_metadata

for retrieving data from the metastore for the location of tensors.

PR #113, opened 2024-07-09
Teacher forcing

Katsutoshii:
Removes autoregressive loop from model.call into model.call_n.
This reduces model memory requirements by a factor of roughly storm_duration / N_FLOOD_MAPS because we no longer predict the entire sequence of flood maps, which unlocks GPU training.


waltaskew:
Let's put together a smaller PR.

This contains more code than we want to review at a time. It also includes some code we know we don't want to merge as is, such as `metastore.py` and `featurelabelchunks.py` (those two files I rewrote as https://github.com/UrbanSystemsLab/climateiq-cnn/pull/112)

Can you put together a PR with just the changes to the model itself to start? The dataset generation we'll want to work on before reviewing & merging.

Katsutoshii:
> Let's put together a smaller PR.
> 
> This contains more code than we want to review at a time. It also includes some code we know we don't want to merge as is, such as `metastore.py` and `featurelabelchunks.py` (those two files I rewrote as #112)
> 
> Can you put together a PR with just the changes to the model itself to start? The dataset generation we'll want to work on before reviewing & merging.

Sure, I've reverted the extra files. The branch with e2e training colab is now at https://github.com/Katsutoshii/climateiq-cnn/tree/teacher-forcing-e2e

usl_models/usl_models/flood_ml/model.py
```diff
@@ -103,80 +111,27 @@ def _validate_and_preprocess_data(
...
+        tb_callback = keras.callbacks.TensorBoard(log_dir=logs,
+                                                     histogram_freq=1,
+                                                     profile_batch='1,100')
+        es_callback = keras.callbacks.EarlyStopping(
+            monitor="val_loss", mode="min", patience=early_stopping
+        )
+        # Fit the model for this sample
+        return self._model.fit(
+            dataset, epochs=epochs, callbacks=[tb_callback] 
+            # + [es_callback] if early_stopping else []
```
waltaskew, line 133: Looks like this is accidentally commented out?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -325,47 +278,54 @@ def __init__(
...
 
         Returns:
             The flood map prediction. A tensor of shape [B, H, W, 1].
         """
+        spatiotemporal: tf.Tensor = input["spatiotemporal"]
+        geospatial: tf.Tensor = input["geospatial"]
+        temporal: tf.Tensor = input["temporal"]
+
+        B = spatiotemporal.shape[0]
+        C = 1
```
tatianawu, line 302: Can we maybe add a comment saying this is the channel dimension for the spatiotemporal tensors? Unlike the other variables, there's no context to infer the variable meaning.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -376,75 +336,56 @@ def forward(
 
         return output
 
-    def call(self, inputs: dict[str, tf.Tensor]) -> tf.Tensor:
+    def call_n(self, full_input: Input, n: int = 1) -> tf.Tensor:
         """Runs the entire autoregressive model.
 
         Args:
             inputs: A dictionary of input tensors.
```
tatianawu, line 343: Need to update the name here to `full_input`. Can you also explain how this differs from the `input` in the `call` method?

usl_models/usl_models/flood_ml/model_params.py
```diff
@@ -23,7 +24,7 @@ class FloodModelParams:
     # Keras parameters.
     # Keras models support both Optimizer objects and (valid) string names.
     # It is the user's responsibility to pass in a valid optimizer value.
-    optimizer: tf.keras.Optimizer | str = "adam"
+    optimizer: Any = dataclasses.field(default_factory=lambda: tf.keras.optimizers.Adam(learning_rate=1e-3))
```
tatianawu, line 27: Will this create a default Optimizer that's passed by reference to all instances if it's not overridden?

I switched to the string value in this [commit](https://github.com/UrbanSystemsLab/climateiq-cnn/pull/93/commits/62c5036ef0a2018337185f123204991e5f4bce25) to avoid possible pass-by-reference issues where the same optimizer gets shared across multiple models, which can cause downstream failures.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -376,75 +336,56 @@ def forward(
...
+            spatiotemporal = tf.concat([spatiotemporal,
+                                        tf.expand_dims(prediction, axis=1)],
+                                       axis=1)[:, 1:]
+
+        # Gather dense tensor out of TensorArray along the time axis.
+        predictions = tf.stack(tf.unstack(predictions.stack()), axis=1)
+        # Drop channels dimension.
+        return tf.squeeze(predictions, axis=-1)
+
+    def _get_temporal_window(self, temporal: tf.Tensor, t: int, n: int) -> tf.Tensor:
```
tatianawu, line 386: Is this a replacement for the `temporal_window_view` function in `data_utils.py`? If so, can you add more detail to the docstring explaining what this is doing? It would also be useful to add a test to make sure the behavior is WAI (and will help others understand what inputs/outputs might look like.

tatianawu (commented):
Did you update the tests based on these changes? Pytest didn't run with the most recent update, but I'd expect some of the tests to no longer be relevant.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -103,80 +112,27 @@ def _validate_and_preprocess_data(
...
+    def call(self, input: Input) -> tf.Tensor:
+        return self._model.call(input)
+
+    def call_n(self, input: Input, n: int = 1) -> tf.Tensor:
+        return self._model.call_n(input, n=n)
+
+    def fit(self, dataset: tf.data.Dataset, epochs: int = 1, early_stopping: int | None = None):
+        """Fit the model to the given dataset."""
+        # Create a TensorBoard callback
+        logs = "./logs/" + datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
```
waltaskew, line 124: what does logging into a 'logs' directory mean in the context of VertexAI training? The logs directory would be in an ephemeral docker container.
This looks like more of a debug option than a default option?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -103,80 +112,27 @@ def _validate_and_preprocess_data(
...
-            epochs=self._model_params.epochs,
-            validation_split=0.2,
-            callbacks=callbacks,
+    def call(self, input: Input) -> tf.Tensor:
+        return self._model.call(input)
+
+    def call_n(self, input: Input, n: int = 1) -> tf.Tensor:
+        return self._model.call_n(input, n=n)
+
+    def fit(self, dataset: tf.data.Dataset, epochs: int = 1, early_stopping: int | None = None):
```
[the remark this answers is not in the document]
Katsutoshii: Correct, I've deprecated that from _model_params to be consistent with keras API.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Sure, I've moved this callback outside of model.fit.

usl_models/usl_models/flood_ml/model_params.py
Katsutoshii: It seems limiting to not be able to specify the parameters of Adam (learning rate etc). Using the default factor avoids these issues as long as the client does not share the Params across multiple models.

tatianawu: The default value doesn't limit anything? It's just there if the user doesn't provide their own optimizer, in which they can specify whatever parameters they want. I have the string as a default to prevent any accidental pass-by-reference issues when the user _doesn't_ specify an optimizer. But your optimizer is equivalent to mine.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Fixed.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -18,6 +20,13 @@
 FloodModelParams: TypeAlias = model_params.FloodModelParams
 
 
+class Input(TypedDict):
```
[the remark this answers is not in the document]
Katsutoshii: I've moved this inside the class, now it's FloodModel.Input.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -103,80 +112,27 @@ def _validate_and_preprocess_data(
...
-            data.labels,
-            batch_size=self._model_params.batch_size,
-            epochs=self._model_params.epochs,
-            validation_split=0.2,
-            callbacks=callbacks,
+    def call(self, input: Input) -> tf.Tensor:
+        return self._model.call(input)
+
+    def call_n(self, input: Input, n: int = 1) -> tf.Tensor:
+        return self._model.call_n(input, n=n)
```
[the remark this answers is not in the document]
Katsutoshii: Updated.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -103,80 +112,27 @@ def _validate_and_preprocess_data(
...
+    def fit(self, dataset: tf.data.Dataset, epochs: int = 1, early_stopping: int | None = None):
+        """Fit the model to the given dataset."""
+        # Create a TensorBoard callback
+        logs = "./logs/" + datetime.datetime.now().strftime("%Y%m%d-%H%M%S")
+        tb_callback = keras.callbacks.TensorBoard(log_dir=logs,
+                                                  histogram_freq=1,
+                                                  profile_batch='1,100')
+        es_callback = keras.callbacks.EarlyStopping(
+            monitor="val_loss", mode="min", patience=early_stopping
+        )
```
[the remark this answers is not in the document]
Katsutoshii: Suggestion applied.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Yes, I've added a unit test for this.

usl_models/tests/__init__.py
waltaskew, line 1: We shoudn't need all the empty `__init__.py`s in python3.
I think I'd actually prefer not to have one here in `usl_models/tests/__init__.py` because it marks it as part of the package installed by `setup.py`.

usl_models/usl_models/flood_ml/dataset.py
```diff
@@ -94,7 +172,8 @@ def generator():
             ),
         ),
     )
-    dataset = dataset.batch(batch_size)
+    if batch_size:
```
waltaskew, line 175: What does this branch protect against? The batch_size being 0?

usl_models/tests/__init__.py
Katsutoshii: I'm on latest pytest (8.2.2) and Python 3.10.14 and the tests are unable to import any of our libraries unless I include the __init__.py files.

usl_models/usl_models/flood_ml/dataset.py
```diff
@@ -16,6 +16,88 @@
 from usl_models.flood_ml import model
 
 
+def load_dataset(
```
waltaskew, line 19: It looks like docstrings & comments are copy-pasted from `load_dataset_windowed`. They reference window-ing, which looks specific to `load_dataset_windowed`.

It would be helpful to explain in the docstrings when a caller should use `load_dataset` vs. `load_dataset_windowed`. Why do we need both?

We should also have a test for this function.

It may be easier to add this in another PR since it doesn't look like this function is used at present.

usl_models/tests/__init__.py
waltaskew: init files in the test modules shouldn't impact the tests' abilities to import other files
it isn't a big deal to include them, but I'm a little curious to understand the issue you're having

usl_models/usl_models/flood_ml/dataset.py
Katsutoshii: Yes, when batch size is zero it avoids adding an extra batch dimension, which was what caused the batch prediction error.

waltaskew: Okay! A comment to that effect would be helpful.

waltaskew: a test for that case would also be helpful so we don't hit this error again!

usl_models/usl_models/flood_ml/dataset.py
Katsutoshii: Fixed.

usl_models/usl_models/flood_ml/model_params.py
```diff
@@ -23,5 +24,17 @@ class FloodModelParams:
...
+
+# Used for testing.
+def test_model_params() -> FloodModelParams:
+    return FloodModelParams(
+        batch_size=4,
+        m_rainfall=3,
+        n_flood_maps=3,
+        lstm_units=32,
+        lstm_kernel_size=3,
+    )
```
tatianawu, line 40: Walt had initially asked for this to be removed from this file and into the `tests/` directory. Is there a reason you moved it back here?

usl_models/tests/flood_ml/flood_model_test.py
```diff
@@ -115,25 +45,23 @@ def test_convlstm_call():
     batch_size = 4
     height, width = 100, 100
     storm_duration = 12
-    params = pytest_model_params()
+    params = model_params.test_model_params()
 
     # The FloodConvLSTM model expects the data to have been preprocessed, such
     # that it receives the full temporal inputs and a single flood map.
```
tatianawu, line 51: I think this comment and the docstring is slightly outdated. Can you update?

usl_models/tests/flood_ml/flood_model_test.py
```diff
@@ -262,3 +146,16 @@ def test_model_checkpoint():
...
+def test_get_temporal_window():
+    """Tests copmuting the temporal window."""
+    B, T_MAX, M = 1, 16, 6
+    temporal = tf.random.normal((B, T_MAX, M))
+
+    t = 1  # Timestep
+    n = 5  # Window size
+    actual = flood_model.FloodConvLSTM._get_temporal_window(temporal, t=t, n=n)
+    assert actual.shape == (B, n, M)
+    expected = tf.concat([np.zeros((B, n - t, M)), temporal[:, 0:t]], axis=1)
```
tatianawu, line 160: This basically seems to just replicate the code from the actual method. Maybe it would be useful to create observable tensors? This has the benefit of also helping users understand what exactly the method is doing.

I'm picturing something similar to https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_models/tests/atmo_ml/atmo_utils_test.py.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -2,10 +2,11 @@
 
 import dataclasses
 import logging
-from typing import Optional, TypeAlias, TypedDict
+from typing import TypeAlias, TypedDict, List, Callable
 
 import tensorflow as tf
-from tensorflow.keras import layers
+import keras
+from keras import layers
```
tatianawu, line 9: Nit: Can we alphabetize?

usl_models/usl_models/flood_ml/model.py
tatianawu: Can you add details to the docstring explaining what the `temporal` tensor looks like when this is called?

Let's add the args and return section; I think it's better to have more detail than less here. The "n-sized window" is a bit confusing, because there are actually two temporal windows of interest in the model: the rainfall `m` and the spatiotemporal `n`. We're actually retrieving the `n` most recent (`m`-sized) rainfall windows at the current time step.

FWIW I wrote the original code and this still trips me up.

usl_models/tests/__init__.py
tatianawu: Did you run `pip install` in your local environment? You'll want:

1. `pip install -r requirements.txt` to make sure you're using the same versions
2. `pip install -e .[dev]` to install the local project

usl_models/tests/flood_ml/flood_model_test.py
Katsutoshii: Done

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/tests/flood_ml/flood_model_test.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model_params.py
Katsutoshii: This was a merge snafu. Reverted.

usl_models/usl_models/flood_ml/dataset.py
Katsutoshii: Separated this into another PR.

usl_models/tests/flood_ml/flood_model_test.py
```diff
@@ -262,3 +157,24 @@ def test_model_checkpoint():
...
+    actual = flood_model.FloodConvLSTM._get_temporal_window(temporal, t=t, n=n)
+    assert actual.shape == (B, n, M)
+    expected = [
+        [
+            np.zeros((M)),
+            np.zeros((M)),
+            np.zeros((M)),
+            [1, 1, 1, 1, 1, 1],
+            [2, 2, 2, 2, 2, 2],
+        ],
```
tatianawu, line 178: Checking my understanding here. With `t = 3`, we have 2 rows of zero padding, and the data retrieved from `temporal` are rows 0-2, correct?

Could we set temporal to 1-8 instead? That row of zeros might be confusing.

tatianawu (approved):
LGTM. I ignored the dataset files. Will all of these be rebased away, or are there any I should take a look at?

SallyElHajjar (approved):

waltaskew (approved):

PR #114, opened 2024-07-10
added solar time and wind dir

MadhaviJain90:
Pls see the main.py code. I might have missed out on some things when adding in the functions for sine and cosine components of wind and solar time! Also changed var name for WSP10, WDIR10 to WSP and WDIR, to avoid confusion with a similar names in wrfouts

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -674,16 +674,13 @@ def _compute_custom_wps_variables(dataset: xarray.Dataset) -> xarray.Dataset:
 
     Returns:
       A new xarray dataset with newly derived variables.
-    """
-    # Derive wind components from UU & VV + Solar time components from XLAT_M & XLONG_M
-   
+    """   
     # WSPD (wind speed) at FNL level 0 (~10m)
     # WDIR_SIN (wind direction_sine component) at FNL level 0 (~10m)
     # WDIR_COS (wind direction_cosine component) at FNL level 0 (~10m)
-    if all(var in dataset.keys() for var in ["UU", "VV", "XLAT_M", "XLONG_M"]):
-        
-        ######## Compute WIND ##########
-
+    
+    if all(var in dataset.keys() for var in ["UU", "VV"]):
+        # Derive wind components from UU and VV: WSPD, WDIR
         uu = dataset.data_vars["UU"]
         vv = dataset.data_vars["VV"]
 
@@ -713,7 +710,8 @@ def _compute_custom_wps_variables(dataset: xarray.Dataset) -> xarray.Dataset:
         wind_direction_sin = direction_to_sine(wind_direction)
         wind_direction_cos = direction_to_cosine(wind_direction)
 
-        # Convert sine and cosine components back to xarray DataArray, retaining original coordinates and dimensions
+        # Convert sine and cosine components back to xarray DataArray
+        # retaining original coordinates and dimensions
         wind_direction_sin = xarray.DataArray(
             wind_direction_sin,
             coords=vv_centered.coords,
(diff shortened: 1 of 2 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -675,24 +675,136 @@ def _compute_custom_wps_variables(dataset: xarray.Dataset) -> xarray.Dataset:
...
+        )
+
+        # Convert solar times to sine and cosine values
+        for utc_time in utc_hours_minutes:
+            sin_solar_time.loc[utc_time, :, :] = time_to_sine(solar_times.sel(utc_time=utc_time))
+            cos_solar_time.loc[utc_time, :, :] = time_to_cosine(solar_times.sel(utc_time=utc_time))
+            
+        new_dims = ["Time", "south_north", "west_east"]
+        dataset = dataset.assign(SOLAR_TIME_SIN=(new_dims, sin_solar_time))
+        dataset = dataset.assign(SOLAR_TIME_COS=(new_dims, cos_solar_time))
```
SallyElHajjar, line 805: Is that the same as what we did yesterday? should i re-check or it is the same?

MadhaviJain90: I added your line of code of fetching time from the file. The rest is my previous code. But I had to make it align with the way the rest of the code was written in main.py. Especially concerned if the way if I have called dataset functions is correct or not @saskatchuwan or @waltaskew pls have a look

usl_pipeline/cloud_functions/main.py
```diff
@@ -675,24 +675,136 @@ def _compute_custom_wps_variables(dataset: xarray.Dataset) -> xarray.Dataset:
...
       A new xarray dataset with newly derived variables.
     """
-    # Derive wind components: WSPD10 (wind speed), WDIR10 (wind direction)
-    if all(var in dataset.keys() for var in ["UU", "VV"]):
+    # Derive wind components from UU & VV + Solar time components from XLAT_M & XLONG_M
+   
+    # WSPD (wind speed) at FNL level 0 (~10m)
+    # WDIR_SIN (wind direction_sine component) at FNL level 0 (~10m)
+    # WDIR_COS (wind direction_cosine component) at FNL level 0 (~10m)
+    if all(var in dataset.keys() for var in ["UU", "VV", "XLAT_M", "XLONG_M"]):
```
saskatchuwan, line 683: I would recommend splitting out the solar time computation, separated from the wind components computation since the solar time will not depend on UU or VV - so like:

```python
# Derive wind components from UU and VV: WSPD, WDIR
if all(var in dataset.keys() for var in ["UU", "VV"]):
    <computation for wind components only>

# Derive solar time from XLAT_M and XLONG_M
if all(var in dataset.keys() for var in ["XLAT_M", "XLONG_M"]):
    <computation for solar time only>
```

saskatchuwan: Up to you but another thing you can do in addition to this to make this code slightly more readable is to break out the wind and solat time computations into helper methods that have descriptive names (to remove need for comments like `#### Compute WIND ####`

example refractor you can do:
```python
def _compute_custom_wps_variables(dataset: xarray.Dataset) -> xarray.Dataset:
     WSPD, WSPD_SIN, WSPD_COS = _compute_wind_components(dataset)
     dataset = dataset.assign(WSPD=(new_dims, wind_speed))
     dataset = dataset.assign(WDIR_SIN=(new_dims, wind_direction_sin))
     dataset = dataset.assign(WDIR_COS=(new_dims, wind_direction_cos))

      SOLAR_TIME_SIN, SOLAR_TIME_COS = _compute_solar_time_components(dataset)
     dataset = dataset.assign(SOLAR_TIME_SIN=(new_dims, sin_solar_time))
     dataset = dataset.assign(SOLAR_TIME_COS=(new_dims, cos_solar_time))

def _compute_wind_components:
    <all the computations for wind>

def _compute_solar_time_components:
    <all the computations for solar time>
```

usl_pipeline/usl_lib/usl_lib/shared/wps_data.py
```diff
@@ -311,5 +322,32 @@ class Unit(Enum):
...
+        # Cosine Component of Solar Time
+        "SOLAR_TIME_COS": {
+            "unit": Unit.NONE,
+            "scaling": {
+                "type": ScalingType.GLOBAL,
+                "min": -1,
+                "max": 1,
+            },
+        },
+        # LU_INDEX
```
saskatchuwan, line 345: Lets move this one next to LANDUSEF (we can do this now bc we haven't generated any labels yet so lets organize it how we like it before it's solidified)

waltaskew:
We should also add some tests for these new features. They should look similar to this:
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/96563734d16d909c6fc872bedb636f5b7ffc6c3f/usl_pipeline/cloud_functions/main_test.py#L529

usl_pipeline/cloud_functions/main.py
```diff
@@ -675,24 +675,136 @@ def _compute_custom_wps_variables(dataset: xarray.Dataset) -> xarray.Dataset:
...
+        ####### Compute Solar Time ##########
+
+        # Extract longitude and time data
+        longitude = dataset.data_vars['XLONG_M']
+        times = dataset.data_vars['Times']
+
+        # Convert time variable to datetime objects
+        times = numpy.array([numpy.datetime64(''.join(t.astype(str)).replace('_', 'T')) for t in times.values])
+        # Extract hours and minutes from the datetime objects
+        utc_hours_minutes = numpy.array([t.astype('datetime64[h]').astype(int) % 24 + t.astype('datetime64[m]').astype(int) % 60 / 60 for t in times])
```
waltaskew, line 743: it looks like we should be able to vectorize these operations, something like
```python
(times.astype('datetime64[h]').astype(int) % 24) + (times.astype('datetime64[m]').astype(int) % 60 / 60)
```

usl_pipeline/usl_lib/usl_lib/shared/wps_data.py
MadhaviJain90: Done! :)

usl_pipeline/cloud_functions/main.py
MadhaviJain90: Maybe for the future. Thanks!!!! :)

PR #117, opened 2024-07-11
Fix for study area chunk metadata initialized too early due to error/retry

rsutormin:
The root cause we're trying to solve in this PR is the following. Before this PR we had a logic avoiding retry for build_feature_matrix CF if the chunk metadata is already present. But if the error happens in this CF we do create chunk metadata object with just error field populated which now blocks retry (it does detect the presence of chunk metadata and stops but it shouldn't in case of previous error).

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -336,7 +336,7 @@ def _build_feature_matrix(
             chunk_metadata = metastore.StudyAreaSpatialChunk.get_if_exists(
                 firestore.Client(), study_area_name, chunk_name
             )
-            if chunk_metadata is not None:
+            if chunk_metadata is not None and chunk_metadata.state is not None:
                 logging.info(
                     "Flood feature matrix for chunk %s was already generated",
                     chunk_path,
@@ -347,6 +347,10 @@ def _build_feature_matrix(
             logging.info(
                 "Start generating flood feature matrix for chunk %s", chunk_path
             )
+            metastore.StudyAreaSpatialChunk(
+                id_=chunk_name,
+                error=firestore.DELETE_FIELD,
+            ).merge(firestore.Client(), study_area_name)
             feature_matrix, metadata = _build_flood_feature_matrix_from_archive(chunk)
 
             if feature_matrix is None:
@@ -844,6 +848,7 @@ def _write_flood_chunk_metastore_entry(chunk_blob: storage.Blob) -> None:
 
     metastore.StudyAreaSpatialChunk(
         id_=chunk_name,
+        state=metastore.StudyAreaChunkState.FEATURE_MATRIX_PROCESSING,
         raw_path=f"gs://{chunk_blob.bucket.name}/{chunk_blob.name}",
         needs_scaling=True,
         x_index=x_index,
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -336,6 +336,13 @@ def _build_feature_matrix(
             chunk_metadata = metastore.StudyAreaSpatialChunk.get_if_exists(
                 firestore.Client(), study_area_name, chunk_name
             )
+            # Let's check if the chunk metadata object is present and has the state
+            # different from None (which means it's either FEATURE_MATRIX_PROCESSING
+            # meaning that unscaled feature matrix is stored and this CF succeeded, or
+            # it's FEATURE_MATRIX_READY and the downstream rescaling CF is also done).
+            # If metadata object is not present it means we're in the first execution
+            # attempt. None state means that we're in the retry and either this CF
+            # crashed during previous execution attempt or it finished with an error.
             if chunk_metadata is not None and chunk_metadata.state is not None:
                 logging.info(
                     "Flood feature matrix for chunk %s was already generated",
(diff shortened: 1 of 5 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -336,7 +336,7 @@ def _build_feature_matrix(
             chunk_metadata = metastore.StudyAreaSpatialChunk.get_if_exists(
                 firestore.Client(), study_area_name, chunk_name
             )
-            if chunk_metadata is not None:
+            if chunk_metadata is not None and chunk_metadata.state is not None:
```
waltaskew, line 339: I think it'd be good to check the value of the state rather than just that it isn't None, e.g. if the state was error, we may want to try again.

Why is this check necessary? 
If the functions are idempotent, it should be safe to re-build the chunk a second time.
If they aren't, I think we'd still have race conditions on checking the state and could still have multiple runners. It would need to be a transactional check & set to ensure one runner at a time.

How does this interact with retries? If one runner sets the state as StudyAreaChunkState.FEATURE_MATRIX_PROCESSING, then dies, would we never be able to re-run?

usl_pipeline/cloud_functions/main.py
```diff
@@ -347,6 +347,10 @@ def _build_feature_matrix(
             logging.info(
                 "Start generating flood feature matrix for chunk %s", chunk_path
             )
+            metastore.StudyAreaSpatialChunk(
+                id_=chunk_name,
+                error=firestore.DELETE_FIELD,
```
waltaskew, line 352: Why do we clear the error here?

waltaskew:
Can you describe the root problem you're trying to solve? Some descriptions in this PR or code comments would be helpful for understanding the changes and what issues they address.

usl_pipeline/cloud_functions/main.py
rsutormin: With my change we have 3 possible values for state:
- None: appears when empty chunk meta object is created few lines below this comment or if the error happened during the unscaled feature matrix building logic in this same CF (but before it finishes with setting the FEATURE_MATRIX_PROCESSING state)
- FEATURE_MATRIX_PROCESSING: CF building unscaled feature matrix ends with this state. If this state appears it means that we no longer need to rerun this CF, it succeeded (no matter if retry logic thinks that there was any deadline exceeded for CF execution time or any other reasons).
- FEATURE_MATRIX_READY: It means that rescaling CF succeeded and we're good to go with this chunk being done completely.

I would love to have INIT state instead of None. But the problem with INIT state is that the logic creating initial empty object shouldn't override existing state that might be set in parallel by CF trying to set FEATURE_MATRIX_PROCESSING for instance. So it means we might need to set INIT state in transaction where we make sure the state is not set yet which is just more complicated logic than keeping the state being None initially.

usl_pipeline/cloud_functions/main.py
rsutormin: Because if we do retry after some error happened in previous attempt to build unscaled matrix, we want to start from clean state again. But I can avoid clearing the error here if you think it's better to keep it till we succeed with building unscaled matrix (we do clear the error at that moment too).

rsutormin:
> Can you describe the root problem you're trying to solve? Some descriptions in this PR or code comments would be helpful for understanding the changes and what issues they address.

Added more details PR description.

How about this explanation in the code? https://github.com/UrbanSystemsLab/climateiq-cnn/pull/117/commits/aaced9bccb383b0e613cb8a02cc92d427fde38be

waltaskew:
> logic avoiding retry for build_feature_matrix CF if the chunk metadata is already present

Is that logic something we need? If the function is idempotent, is it okay to rebuild the matrix again?
Could that logic prevent us from performing re-runs that we do want -- re-runs due to code changes, failure retries, etc?

rsutormin:
> > logic avoiding retry for build_feature_matrix CF if the chunk metadata is already present
> 
> Is that logic something we need? If the function is idempotent, is it okay to rebuild the matrix again? Could that logic prevent us from performing re-runs that we do want -- re-runs due to code changes, failure retries, etc?

Yeah, I added that check as a result of a problem when we saw build_feature_matrix CF succeeded but got retried because it was just a bit over the 1 minute threshold. The retry loop produced a lot of events that triggered many rescaling passes. That was last needed chunk (longest to process) that was needed to start rescaling so with it being done in every retry rescaling logic started full check again and again.

This check will prevent re-runs due to code changes - in this case we would have to re-upload from scratch forcing initial cleanup in uploader. But this wouldn't prevent failure retries since chunk metadata state update is the last action of the CF building unscaled feature matrix (https://github.com/UrbanSystemsLab/climateiq-cnn/blob/96563734d16d909c6fc872bedb636f5b7ffc6c3f/usl_pipeline/cloud_functions/main.py#L395).

usl_pipeline/cloud_functions/main.py
waltaskew: I think it'd be good to keep the error around until we actually succeed in case the next invocation is unable to write something out to avoid losing data.

waltaskew (approved):
This is a hard problem! Thanks for explaining and adding code comments.
This might be a good thing for us to spend some time brainstorming about.

My bias is that if a retry happens we should do it rather than try to figure out if it was a retry that should happen. Which is essentially what we're doing here -- we're really second-guessing the retry invocation to see if we think it was a bad retry.

I think it would be more direct to try and control our retries better so we aren't in a situation where we have to guess about the validity of a retry.

PR #118, opened 2024-07-11
[WRF] Bug fixes for build_feature_matrix for wps

saskatchuwan:
* Update parsing for timestamp from netfdf when writing to metastore - needs to be decoded (similar to what is done for labels: https://github.com/UrbanSystemsLab/climateiq-cnn/pull/97)
* Fix chunk_path reference so it is accessible by wrf's sectino of the if/else statement


waltaskew (approved):
LGTM

PR #120, opened 2024-07-14
[WRF] Update build_wrf_label_matrix to trigger metastore write for `Simulation` entry

saskatchuwan:
* Write `Simulation` entry to metastore when building wrf label matrix
    * Simulation entry will contain reference to a corresponding `HeatScenarioConfig` introduced in: https://github.com/UrbanSystemsLab/climateiq-cnn/pull/149

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -381,50 +381,6 @@ def delete_flood_scenario_metadata(cloud_event: functions_framework.CloudEvent) 
 
 @functions_framework.cloud_event
 @_retry_and_report_errors()
-def build_and_upload_simulation_chunk(
-    cloud_event: functions_framework.CloudEvent,
-) -> None:
-    """Creates a simulation chunk and uploads it to GCS.
-
-    This function is triggered when files containing raw output data for a simulation
-    is uploaded to GCS. It will load the file into simluation chunks bucket and
-    write a new simluation entry to metastore.
-    """
-    file_name = pathlib.PurePosixPath(cloud_event.data["name"])
-    bucket_name = cloud_event.data["bucket"]
-
-    # For AtmoML, only 3rd domain (500m) output files will be used
-    if re.search(file_names.WRF_DOMAIN3_NC_REGEX, file_name.name):
-        storage_client = storage.Client()
-        db = firestore.Client()
-
-        bucket = storage_client.bucket(bucket_name)
-        file_blob = bucket.blob(str(file_name))
-        blob_path = pathlib.PurePosixPath(cloud_event.data["name"])
-
-        # For WRF, we can treat each snapshot output file as its own chunk. If
-        # multiple snapshots must be processed together, we can choose to group
-        # the files together in a TAR before uploading.
-        with file_blob.open("rb") as fd:
-            simulation_chunk_bucket = storage_client.bucket(
-                cloud_storage.SIMULATION_CHUNKS_BUCKET
-            )
-            simulation_chunk_bucket.blob(str(file_name)).upload_from_file(fd)
-
-            # File names should be in the form <study_area_name>/<file_name>
-            study_area_name = file_name.parent.name
-
-            # Enter the simulation in our metastore.
-            metastore.Simulation(
-                gcs_prefix_uri=f"gs://{file_blob.bucket.name}/{blob_path.parent}",
-                simulation_type=metastore.SimulationType.WRF,
-                study_area=metastore.StudyArea.get_ref(db, study_area_name),
-                configuration=None,
-            ).set(db)
-
-
-@functions_framework.cloud_event
-@_retry_and_report_errors()
 def build_wrf_label_matrix(cloud_event: functions_framework.CloudEvent) -> None:
     """Builds a label matrix when a set of simulation output files is uploaded.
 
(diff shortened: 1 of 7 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -329,6 +329,50 @@ def delete_flood_scenario_metadata(cloud_event: functions_framework.CloudEvent)
...
+
+            # File names should be in the form <study_area_name>/<file_name>
+            study_area_name = file_name.parent.name
+
+            # Enter the simulation in our metastore.
+            metastore.Simulation(
+                gcs_prefix_uri=f"gs://{file_blob.bucket.name}/{blob_path.parent}",
+                simulation_type=metastore.SimulationType.WRF,
+                study_area=metastore.StudyArea.get_ref(db, study_area_name),
+                configuration=None,
```
waltaskew, line 372: Why aren't these associated with a configuration? I'd think they be associated with some configuration like '90-th percentile NY summer' or such

usl_pipeline/cloud_functions/main.py
```diff
@@ -329,6 +329,50 @@ def delete_flood_scenario_metadata(cloud_event: functions_framework.CloudEvent)
...
+        blob_path = pathlib.PurePosixPath(cloud_event.data["name"])
+
+        # For WRF, we can treat each snapshot output file as its own chunk. If
+        # multiple snapshots must be processed together, we can choose to group
+        # the files together in a TAR before uploading.
+        with file_blob.open("rb") as fd:
+            simulation_chunk_bucket = storage_client.bucket(
+                cloud_storage.SIMULATION_CHUNKS_BUCKET
+            )
+            simulation_chunk_bucket.blob(str(file_name)).upload_from_file(fd)
```
waltaskew, line 362: It looks like this is a straight copy of the file?
[copy_blob](https://gcloud.readthedocs.io/en/latest/storage-buckets.html#google.cloud.storage.bucket.Bucket.copy_blob)
should be able to do this more cheaply.

I'd also ask if we could upload directly to the simulations bucket rather than leave two copies of the same file in different buckets.

usl_pipeline/cloud_functions/main.py
saskatchuwan: I thought that the configs were flood specific? I don't think the wrf team knows about the configs - when they kick off their wrf sim runs, they will do it directly in the hpc cluster

usl_pipeline/cloud_functions/main.py
saskatchuwan: Ack  on the two copies - I have it set up this way to mirror what we're doing for the [features](https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_pipeline/cloud_functions/main.py#L151) to make it more predictable (https://miro.com/app/board/uXjVNgvVs2Q=/)

The other option if we want to just keep one copy of the file is to just write the Simulation entry when we [build the wrf label matrix](https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_pipeline/cloud_functions/main.py#L335). WDYT

usl_pipeline/cloud_functions/main.py
waltaskew: If we don't store literal WRF configs that's okay, but we should have some sort of indication of what the simulation data describes. If we have multiple simulations for the same study area, we'll want to be able to tell the simulations apart. So we'd want some marker describing the configuration which produced the simulation.

usl_pipeline/cloud_functions/main.py
waltaskew: creating the entry when we create the label matrix sounds good to me

usl_pipeline/cloud_functions/main.py
saskatchuwan: I see - i think the differentiation would be the month+year that we're running for (for ex: 2010 summer vs 2013 summer)

@MadhaviJain90 Do you have any thoughts here?

saskatchuwan: @waltaskew in `SImulation`, it expects Configuration to be an existing Firestore document. We won't have that for WRF. Once way to bake in a differentiating SImulation ID between each wrf run is to do have a different `set` function that takes in a manually constructed Simulation ID instead of impliciting building it from <study_are>-<configuration>

So for WRF - the simulation id's could be like:
1. `<study_area>-<time_range>` = NYC_Heat-201006
2. `<study_area>-<resoluation>-<time_range>` = NYC_Heat-500m-201306 

@MadhaviJain90 WDYT?

waltaskew: In the flood context, the configuration is a rainfall pattern which has its own tensor representation that is fed into the model. So that tensor is what tells the model to predict a 1 or 100 year flood.

I'd assume we'd need something similar for WRF -- we'd have some tensor representing the prevailing atmospheric conditions? Something to tell the model it should be predicting a hot summer?

But if we don't have such (right now?) that's fine -- the configuration would be more informational and just state 'which simulation is this' rather than a description of a tensor that would be fed into the model. You could create that configuration document right here at the same time you create the simulation document.

Using that configuration document to state the time, place & resolution sounds good to me.

saskatchuwan: Thanks for the input here's what I'm thinking:
1. Create new `climateiq-atmospheric-sim-config` bucket and create new CF for it that will expect a Heat_Config.txt
    * This config file will be created and stored in this bucket by the WRF team to contain data like: `sim_year`, `percentile`
    * CF will write new HeatScenarioConfig to metastore
2. Keep build_wrf_label as is - it will write a new `Simulation` metastore entry that will look for the above config

saskatchuwan: @MadhaviJain90 @waltaskew ^^

saskatchuwan: Ack - I don't think we need city since we'll always have the city attached to the study area we can have the config include:
```
simulation_year=2012
simulation_start_month=6
simulation_end_month=8
percentile=99
```

WDYT?

saskatchuwan:
@waltaskew Please take another look! Updated the `Simulation` entry for wrf to reference and store path to its corresponding config (introduced: https://github.com/UrbanSystemsLab/climateiq-cnn/pull/149)

PR #123, opened 2024-07-15
Updates to dataset.py

Katsutoshii:
Updates dataset.py to include channels dimension for spatiotemporal tensors and add `load_dataset` implementation.

### DIFF
--- a/usl_models/tests/flood_ml/test_dataset.py
+++ b/usl_models/tests/flood_ml/test_dataset.py
@@ -9,6 +9,121 @@ from usl_models.flood_ml import dataset
 
 
 @mock.patch.object(dataset, "metastore")
+def test_load_dataset_full(mock_metastore) -> None:
+    """Ensures we create the expected dataset from GCS objects."""
+    mock_firestore_client = mock.MagicMock(spec=firestore.Client)
+    mock_storage_client = mock.MagicMock(spec=storage.Client)
+
+    batch_size = 1
+
+    # Set some URLs to return from our metastore functions.
+    mock_metastore.get_temporal_feature_metadata.return_value = {
+        "as_vector_gcs_uri": "gs://temporal-features/temporal-feature.npy"
+    }
+    mock_metastore.get_spatial_feature_and_label_chunk_metadata.return_value = [
+        (
+            {"feature_matrix_path": "gs://spatial-features/spatial-feature.npy"},
+            {"gcs_uri": "gs://labels/labels.npy"},
+        )
+    ] * batch_size
+
+    # Set up some fake data to retrieve from our mock GCS.
+    mock_rainfall = numpy.pad(numpy.array([1, 2, 3, 4]), (0, 860))
+    mock_spatial_features = numpy.random.rand(1000, 1000, 8)
+    mock_labels = numpy.random.rand(1000, 1000, 19)
+
+    # Set the storage client to return the right mock data for each GCS path.
+    def mock_blob_func(blob_name):
+        """Returns the mock data depending for blob requested."""
+        if blob_name == "temporal-feature.npy":
+            mock_data = mock_rainfall
+        elif blob_name == "spatial-feature.npy":
+            mock_data = mock_spatial_features
+        elif blob_name == "labels.npy":
+            mock_data = mock_labels
+        else:
+            raise ValueError(f"Unexpected name {blob_name} passed to mock.")
+
+        buf = io.BytesIO()
+        numpy.save(buf, mock_data)
+        buf.seek(0)
+
+        mock_blob = mock.MagicMock(spec=storage.Blob)
+        mock_blob.open.return_value = buf
+        return mock_blob
+
+    mock_storage_client.bucket().blob.side_effect = mock_blob_func
+    mock_storage_client.reset_mock()
+
+    m_rainfall = 6
+    ds = dataset.load_dataset(
+        sim_names=["sim_name"],
+        batch_size=batch_size,
+        n_flood_maps=5,
+        m_rainfall=m_rainfall,
+        firestore_client=mock_firestore_client,
+        storage_client=mock_storage_client,
+    )
+    batch = ds.take(1)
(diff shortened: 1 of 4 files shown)


usl_models/usl_models/flood_ml/dataset.py
```diff
@@ -16,6 +16,87 @@
 from usl_models.flood_ml import model
 
 
+def load_dataset(
```
waltaskew, line 19: Could you describe in the docstrings when & why a caller should use `load_dataset` vs. `load_dataset_windowed` ?

usl_models/usl_models/flood_ml/dataset.py
```diff
@@ -16,6 +16,87 @@
...
+                    dtype=tf.float32,
+                ),
+            ),
+            tf.TensorSpec(
+                shape=(None, constants.MAP_HEIGHT, constants.MAP_WIDTH),
+                dtype=tf.float32,
+            ),
+        ),
+    )
+    if batch_size:
```
waltaskew, line 95: Can you add a comment explaining why we need a special case for a batch_size of 0?

usl_models/tests/flood_ml/test_dataset.py
```diff
@@ -86,45 +86,48 @@ def mock_blob_func(blob_name):
...
-                    [
-                        numpy.zeros((1000, 1000)),
-                        numpy.zeros((1000, 1000)),
-                        mock_labels[:, :, 0],
-                        mock_labels[:, :, 1],
-                        mock_labels[:, :, 2],
-                    ]
-                ),
-            ]
+        numpy.expand_dims(
```
waltaskew, line 89: Can you write out the tensor we're expecting rather than call `expand_dims` here? It's hard to tell what we're testing if the test code mirrors the calls in the code being tested to produce the expected outputs.

usl_models/usl_models/flood_ml/dataset.py
```diff
@@ -16,6 +16,87 @@
...
+                n_flood_maps,
+                m_rainfall,
+                max_chunks,
+            ):
+                yield (model_input, labels)
+
+    # Create the dataset for this simulation
+    dataset = tf.data.Dataset.from_generator(
+        generator=generator,
+        output_signature=(
```
waltaskew, line 65: it looks like this is the same signature as below -- is it possible to factor out a function for the duplication?

usl_models/tests/flood_ml/test_dataset.py
```diff
@@ -86,45 +86,48 @@ def mock_blob_func(blob_name):
     # geospatial will have the labels creeping into a sequence of zeros.
```
waltaskew, line 86: It'd be good to add a test for `load_dataset`

waltaskew:
It might be faster to just merge the fix to `load_dataset_windowed` in this PR and worry about adding `load_dataset` later. Do we need `load_dataset` over `load_dataset_windowed` right now?

usl_models/usl_models/flood_ml/dataset.py
Katsutoshii: Done

usl_models/tests/flood_ml/test_dataset.py
Katsutoshii: Done.

usl_models/tests/flood_ml/test_dataset.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/dataset.py
Katsutoshii: Temporal tensor and label tensor are different. Probably not worth factoring it out, then it becomes more complex to read.

usl_models/usl_models/flood_ml/dataset.py
Katsutoshii: Done.

Katsutoshii:
> It might be faster to just merge the fix to `load_dataset_windowed` in this PR and worry about adding `load_dataset` later. Do we need `load_dataset` over `load_dataset_windowed` right now?

load_dataset is important for testing call_n feature, so I would rather we get that checked in ASAP.

usl_models/usl_models/flood_ml/dataset.py
```diff
@@ -16,6 +16,91 @@
...
+    batch_size: int = 4,
+    n_flood_maps: int = constants.N_FLOOD_MAPS,
+    m_rainfall: int = constants.M_RAINFALL,
+    max_chunks: Optional[int] = None,
+    firestore_client: Optional[firestore.Client] = None,
+    storage_client: Optional[storage.Client] = None,
+) -> tf.data.Dataset:
+    """Creates a dataset which generates chunks for flood model inference.
+
+    For training with teacher-forcing, `load_dataset_windowed` should be used instead.
```
waltaskew, line 30: I think it'd be helpful to describe when to use this function, as opposed to when not to use it as we do here.

usl_models/usl_models/flood_ml/dataset.py
```diff
@@ -16,6 +16,91 @@
...
+                    dtype=tf.float32,
+                ),
+            ),
+            tf.TensorSpec(
+                shape=(None, constants.MAP_HEIGHT, constants.MAP_WIDTH),
+                dtype=tf.float32,
+            ),
+        ),
+    )
+    # If no batch specified, do not batch the dataset, which is required
```
waltaskew, line 96: Is the batch_size `None` in that case? Or is it 0?
If `None` it'd be good to do
```python
batch_size: Optional[int] = 4,
```
above.
And also have 
```python
if batch_size is not None
```
here

waltaskew (approved):
LGTM, just some small things

usl_models/usl_models/flood_ml/dataset.py
Katsutoshii: We've been using batch_size = 0.

PR #124, opened 2024-07-16
Check in training notebook

Katsutoshii:
Check in training notebook, remove print statements.


usl_models/train_flood_model.ipynb
```diff
@@ -0,0 +1,151 @@
+{
```
waltaskew, line 1: let's stick this someplace like a directory named 'examples' rather than at the package root

waltaskew (commented):
let's stick this someplace like a directory named 'examples' rather than at the package root

Katsutoshii:
> let's stick this someplace like a directory named 'examples' rather than at the package root

Done, moved to `usl_models/notebooks` folder.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
waltaskew, line 1: looks like this was accidentally included

waltaskew (approved):

PR #126, opened 2024-07-16
[WRF] Fix metastore bugs from feature matrix pipeline

saskatchuwan:
1. Bug: Firestore document id restrict use of certain chars
    * Fix: Sanitize doc ids - replace all "." and ":" from chunk name with underscore before using it as document id
2. Bug: In `build_and_upload_study_area_chunk()`, nc file could not be opened from python stream due to invalid arg
   * Fix: Since stream was first used in GCS upload before being read again to create in memory dataset, it needs to be seeked to beginning
3. Bug: metastore.StudyArea failed to persist x_ll_corner and y_ll_corner due to invalid type
    * Fix: Convert values from numpy.float32 -> float

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -1026,8 +1026,11 @@ def _write_wps_chunk_metastore_entry(
     db = firestore.Client()
     study_area_name, chunk_name = _parse_chunk_path(chunk_blob.name)
 
+    # Remove unsupported characters for firestore
+    doc_id = chunk_name.replace(".", "_").replace(":", "_")
+
     metastore.StudyAreaTemporalChunk(
-        id_=chunk_name,
+        id_=doc_id,
         raw_path=f"gs://{chunk_blob.bucket.name}/{chunk_blob.name}",
         feature_matrix_path=f"gs://{feature_blob.bucket.name}/{feature_blob.name}",
         time=metadata.time,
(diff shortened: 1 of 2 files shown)


saskatchuwan:
@waltaskew / @rsutormin pls take a look at some fixes from e2e testing yesterday. wiould like to deploy this today and verify fixes in dev. thanks!

usl_pipeline/cloud_functions/main.py
```diff
@@ -180,15 +180,17 @@ def build_and_upload_study_area_chunk(
             # File names should be in the form <study_area_name>/<file_name>
             study_area_name = file_name.parent.name
 
+            fd.seek(0)  # Seek to the beginning so the file can be read.
```
waltaskew, line 183: it looks like we're copying the file unchanged between GCS locations, so another way to handle this would be to use [copy_blob](https://gcloud.readthedocs.io/en/latest/storage-buckets.html#google.cloud.storage.bucket.Bucket.copy_blob) rather than `upload_from_file(fd)`. It should also be cheaper.

waltaskew (approved):

usl_pipeline/cloud_functions/main.py
saskatchuwan: SG I'll work on replacing it with copy as an optimiziation. added a TODO

PR #127, opened 2024-07-16
add signature for call_n when saving

waltaskew:


waltaskew:
we need to make some more changes to `call_n` before it can actually be serialized -- for instance, it will need to drop the Python `for` for a `tf.while`

PR #129, opened 2024-07-16
Scripts for VertexAI Training

waltaskew:


prashantkul (approved):

usl_models/trainer/flood_task.py
```diff
@@ -0,0 +1,133 @@
...
+    # non-primary workers for multi-workers
+    else:
+        # each worker saves their model instance to a unique temp location
+        worker_dir = args.model_dir + "/workertemp_" + str(task_id)
+        logging.info("Saving model to %s", worker_dir)
+        tf.io.gfile.makedirs(worker_dir)
+        model.save_model(worker_dir)
+
+
+with strategy.scope():
```
Katsutoshii, line 115: nit: should this be guarded behind __name__ == "__main__"?

usl_models/trainer/flood_task.py
```diff
@@ -0,0 +1,133 @@
...
+# Single Machine, single compute device
+if args.distribute == "single":
+    if tf.test.is_gpu_available():
+        strategy = tf.distribute.OneDeviceStrategy(device="/gpu:0")
+    else:
+        strategy = tf.distribute.OneDeviceStrategy(device="/cpu:0")
+    logging.info("Single device training")
+# Single Machine, multiple compute device
+elif args.distribute == "mirrored":
+    strategy = tf.distribute.MirroredStrategy()
```
Katsutoshii, line 54: Have we tried this strategy yet? If this works we should definitely use it for the larger training job, we could support a larger batch size.

Katsutoshii (approved):

usl_models/trainer/flood_task.py
waltaskew: I have tried none of these strategies -- this is stuff I copy-pasted from vertex documentation but haven't actually tried.

usl_models/trainer/flood_task.py
waltaskew: I don't think so -- this isn't running as a script. I think it's getting imported by Vertex stuff which causes it to be run rather than run as a script -- the interface has you passing a module name

PR #132, opened 2024-07-17
[WRF] Add missing `cartopy` package in requirements.txt

saskatchuwan:
Fixes bug when CF attempts to write to metastore and utilizes cartopy to extract CRS value

### DIFF
--- a/usl_pipeline/cloud_functions/requirements.txt
+++ b/usl_pipeline/cloud_functions/requirements.txt
@@ -2,6 +2,7 @@ affine==2.4.0
 attrs==23.2.0
 blinker==1.7.0
 cachetools==5.3.3
+cartopy==0.23.0
 certifi==2024.2.2
 charset-normalizer==3.3.2
 click==8.1.7


usl_pipeline/cloud_functions/requirements.txt
```diff
@@ -2,6 +2,7 @@ affine==2.4.0
 attrs==23.2.0
 blinker==1.7.0
 cachetools==5.3.3
+cartopy==0.23.0
```
waltaskew, line 5: are there no addition dependencies-of-dependencies to include?

waltaskew (approved):

usl_pipeline/cloud_functions/requirements.txt
saskatchuwan: shoot missed this - for wrf python in particular?

I could do a pip freeze as a follow up. we just want to exclude the packages we're using for development purposes rihgt?

waltaskew: right, I usually do
```bash
rm venv
virtualenv venv
source venv/bin/activate
pip install -r requirements.txt <my-new-package>
pip freeze > requirements.txt
```

PR #133, opened 2024-07-17
Add some preliminary evaluation metrics

tatianawu:


usl_models/usl_models/flood_ml/eval.py
```diff
@@ -0,0 +1,46 @@
...
+        labels: [N, H, W] array of ground-truth labels.
+
+    Returns:
+        [H, W] array of the absolute error, averaged over the N examples.
+    """
+    return np.mean(np.absolute(predictions - labels), axis=0)
+
+
+def time_series_errors(
+    predictions: np.ndarray, labels: np.ndarray
```
waltaskew, line 20: Should these be defined in terms of tf.Tensors instead of numpy arrays? 
When training, we'll have access to tf.datasets for the labels and the prediction results I'd assume would be a tf.Tensor. So it'd be helpful to define these in terms of datasets & labels.

tatianawu: I think the code should still work on tensors, since they're array-like objects. It looks like [predict](https://www.tensorflow.org/api_docs/python/tf/keras/Model#predict) returns numpy array(s), which is why I defined them this way. Is tf.dataset different, where it wouldn't work with this code?

In general, I'd imagine these functions are getting called _after_ we have a fully-trained model and want to do a more comprehensive evaluation on the test set. During model training, I think it's more common to just look at validation metrics, since looking at test metrics could lead to bias in experiments.

waltaskew: I'd imagined doing a fit() on the holdout set at the end of training to calculate some metrics, then record said metrics when we save the model.

tatianawu: Oh, that's a definite no. We should never train on the test set.

waltaskew: Sorry, I meant 'predict' above, not fit.

I'm suggesting that the training script would do
```python
train, test = get_dataset()
model.fit(train)
predictions = model.predict(test)
metric = get_metrics(predictions, test)
save_model(model, metrics)
```

tatianawu: Ah okay, that should be fine. What exactly does `get_dataset()` return? I'm expecting an array-like object (np.array, tf.Tensor).

waltaskew: it's a tf.data.Dataset

the thing we get from 
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/c78df879397c3f5395a5fed264fdd384d2e49043/usl_models/usl_models/flood_ml/dataset.py#L19

which we can then call fit on
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/c78df879397c3f5395a5fed264fdd384d2e49043/usl_models/trainer/flood_task.py#L128
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/c78df879397c3f5395a5fed264fdd384d2e49043/usl_models/trainer/flood_task.py#L87

tatianawu: I don't really understand how Dataset works and what we'd need to do on top of it. But I think these functions should be general enough to work outside of the model, if I were to just pass in 2 arrays.  So I think it's up to the model to create `test` in a way that's acceptable by this code, not the other way around.

tatianawu: I find it a bit strange that we want to run evaluation scripts during training like this. I'm more used to running evaluation on a "final model" after training. For example, I don't think it makes sense to calculate these metrics during experiments and hyperparameter tuning, since it might result in bias.

I'd envisioned we do something like:
1. Full training
2. Predict on the test set
3. Run evaluation on those predictions + labels

@Katsutoshii I think you probably have the best understanding of both ML + TensorFlow mechanics. Do you know what we might need to do to pass a `tf.Dataset` in to these functions? I'd very much prefer to leave this code as is and not deal with datasets here, if possible.

Katsutoshii: To my knowledge, the canonical way to do this is to follow https://www.tensorflow.org/guide/keras/training_with_built_in_methods#custom_metrics.
`model.evaluate` should be able to take a `tf.data.Dataset` without any modification, provided that your methods are valid TF functions: https://www.tensorflow.org/guide/keras/training_with_built_in_methods#training_evaluation_from_tfdata_datasets.
If we have no reason not to, I vote we use `tf` tensors and ops. They're strictly more flexible for what we're trying to do.

waltaskew: > I find it a bit strange that we want to run evaluation scripts during training like this. I'm more used to running evaluation on a "final model" after training.

I think that this is the final model after training. I'd want to include all the training activities in this script rather than have another process take over.

tatianawu: So I'm not talking about an arbitrary model after training. I'm talking about a "final model" after running experiments and hyperparameter tuning, once we feel comfortable using it to generate our frontend predictions (i.e., the best model we can achieve). Maybe I'm approaching this from a different angle. I had in mind that these are evaluating our _predictions_, not the model. I think I can implement these equivalently with TF ops, as Josiah suggested, but I think it would be useful to clarify use cases.

What about if we want to call these without training? For example, say we have a trained NYC model and want to test it out on Phoenix before we train a new model to see how it generalizes. We'd only need to call prediction, not training, but we'd still want to calculate these evaluation metrics. Is the data in GCS going to be TF tensors or numpy arrays?

waltaskew (approved):

usl_models/usl_models/flood_ml/eval.py
tatianawu: Updated to expect tf.Tensors instead.

PR #134, opened 2024-07-18
Train sampling

Katsutoshii:
Instead of a sliding window, each batch now selects a random (simulation, chunk, timestep) per example to improve diversity per batch.


waltaskew (commented):
This PR has `train-notebook` as its merge base rather than `main`, which I think it should be.

PR #137, opened 2024-07-18
[WRF] Ensure nested directories are handled properly in metastore writes

saskatchuwan:
Within a study for WRF, we may use subdirectories to further organize output files that come from different sim runs. Currently, if we have an output file in a nested subfolder, a new StudyArea within the metastore would be created named after that subfolder. Instead, we want to ensure that:
1. Root folder determines `StudyArea`
2. StudyAreaChunks are always written in metastore like `<study_area>/chunks/<chunk_name>` regardless of folder organization in source bucket destination

GCS structure
```
/NYC_Heat
    /Summer_2010
         met_em1.nc
    /Summer_2015
         met_em20.nc
   ....
```
should result in

Metastore structure:
```
NYC_Heat            (StudyArea)
    /chunks
        met_em1      (StudyAreaChunk)
        met_em20   (StudyAreaChunk)
        ...
```

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -179,8 +179,9 @@ def build_and_upload_study_area_chunk(
             # copy
             study_area_chunk_bucket.blob(str(file_name)).upload_from_file(fd)
 
-            # File names should be in the form <study_area_name>/<file_name>
-            study_area_name = file_name.parent.name
+            # File names can be in the form <study_area_name>/<file_name>
+            # or <study_area_name>/<some_sub_folder>/<file_name>
+            study_area_name = file_name.parts[0]
 
             fd.seek(0)  # Seek to the beginning so the file can be read.
             nc_bytes = fd.read()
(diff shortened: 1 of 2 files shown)


saskatchuwan:
@MadhaviJain90 @waltaskew Cna you review the structure of the metastore data? This will probably be the last thing we should fix in the features portion of the pipeline. This will allow wrf team to do their own file organization in GCS without it affecting the organization of the metastore entries

waltaskew (approved):
LGTM

saskatchuwan:
> Looks great. Although we could have a nested directory naming convention as - Season_YYYY_xxp e.g. Summer_2010_99p. xx being the percentile. This shouldn't affect the code execution, right? I want to minimize having to look for separate docs to confirm which WRF run corresponds which percentile and year

Yeah that's fine the naming of the subfolders can be whatever you think will be the most informative - we won't use the names of the subfolders for the metastore. It's only the name of the root folder that needs to follow a specific naming convention. I've documented these details in the atmoML data pipeline doc in notion. So take a look and feel free to leave feedback.comments there!

MadhaviJain90 (commented):
Looks good

PR #138, opened 2024-07-18
Add a validation split to label chunks

tatianawu:
Add a validation set. This is now a 60/20/20 train/val/test split, indicated by the `dataset` flag in the metastore.


waltaskew (approved):
LGTM - only suggestion I can think of would be to use an enum rather than a string for the three dataset names.

tatianawu:
> LGTM - only suggestion I can think of would be to use an enum rather than a string for the three dataset names.

Updated to use an enum instead.

PR #139, opened 2024-07-19
Write Metadata for Trained Models to the Metastore

waltaskew:


usl_models/usl_models/flood_ml/metastore.py
```diff
@@ -1,8 +1,42 @@
...
+    Returns:
+      The document ID of the written model metadata document.
+    """
+    # Use the GCS location of the model as a unique value for the ID.
+    # URL-escape the ID, as characters such as / are not allowed in document IDs.
+    model_id = urllib.parse.quote(gcs_model_dir, safe=())
+    db.collection("models").document(model_id).set(
+        {
+            "trained_on": [_get_simulation_doc(db, sim_name) for sim_name in sim_names],
+            "trained_at_utc": datetime.datetime.utcnow(),
```
tatianawu, line 33: What's the value of this? Is this mostly to track when we saved a model?

usl_models/usl_models/flood_ml/metastore.py
```diff
@@ -1,8 +1,42 @@
+import datetime
 from typing import Any, Sequence, TypeVar
 import urllib.parse
 
 from google.cloud import firestore  # type:ignore[attr-defined]
 
+import usl_models.flood_ml.model_params
+
+
+def write_model_metadata(
```
tatianawu, line 10: Do any of the current fields hold information about the number of epochs we trained on? If not, I think that could be useful to add too.

usl_models/usl_models/flood_ml/metastore.py
waltaskew: yeah, to help distinguish the different experiments -- a better marker would be something like a commit hash or a version of the code trained on

usl_models/usl_models/flood_ml/metastore.py
waltaskew: the epochs are part of the model params, which we write out

Screenshot 2024-07-19 3 11 11 PM

waltaskew: oh, wait, I'm wrong -- they used to be part of the params but no longer are! I'll add that in!

usl_models/usl_models/flood_ml/metastore.py
tatianawu: What about adding something like an experiment name or description? So the trainer task could (optionally) take in a short descriptor and include that when we save the model?

usl_models/usl_models/flood_ml/metastore.py
tatianawu: I think we're missing the field in the docstring.

tatianawu (approved):

usl_models/usl_models/flood_ml/metastore.py
waltaskew: now with model name!

PR #140, opened 2024-07-19
Set the Default Batch Size to 4

waltaskew:
- That's all that will fit in memory!

### DIFF
--- a/usl_models/usl_models/atmo_ml/model_params.py
+++ b/usl_models/usl_models/atmo_ml/model_params.py
@@ -24,7 +24,7 @@ class AtmoModelParams(TypedDict):
 def default_params() -> AtmoModelParams:
     """Creates default model parameter values."""
     return {
-        "batch_size": 64,
+        "batch_size": 4,
         "lstm_units": 512,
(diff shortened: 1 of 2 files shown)


usl_models/usl_models/atmo_ml/model_params.py
```diff
@@ -24,7 +24,7 @@ class AtmoModelParams(TypedDict):
 def default_params() -> AtmoModelParams:
     """Creates default model parameter values."""
     return {
-        "batch_size": 64,
+        "batch_size": 4,
```
tatianawu, line 27: Do we know that this applies to atmo as well? I think the inputs are (200, 200), though we do have more features in this model.

I'm fine with changing it. USL can always change this later if they end up able to train on larger batches.

tatianawu (approved):

usl_models/usl_models/atmo_ml/model_params.py
waltaskew: I have no clue for atmo -- just guessing it'll be closer to 4 than 64 given our flood struggles

PR #144, opened 2024-07-20
Improve Documentation for Study Area Uploader

waltaskew:
- Add more README information.
- Allow the main.py script to be run directly.
- Change name of argument from 'override' to 'overwrite'
- Remove the chunk size option, as any value besides 1000 will result in issues.
- Provide additional information in the help for arguments.
- Group similar CLI args together.


usl_pipeline/study_area_uploader/study_area_uploader/main.py
```diff
@@ -108,28 +108,29 @@ def main() -> None:
             prepared_inputs,
             work_dir,
             chunk_bucket,
-            args.chunk_length,
+            1000,
```
rsutormin, line 111: Oh, thanks for catching this!

rsutormin (approved):

PR #145, opened 2024-07-20
changes to support prediction dataset

prashantkul:
Folks,

Please review these changes to support prediction dataset.  You should see output like this:

```study_area = "NYC"
city_cat_config = "config_v1%2FRainfall_Data_1.txt"
dataset = load_prediction_dataset(study_area=study_area, city_cat_config=city_cat_config, batch_size=1, max_chunks=1)
input_tuple = next(iter(dataset))

model_input, metadata = input_tuple

print("Model Input:")
for key, value in model_input.items():
    print(f"Key: {key}")
    print(f"Shape: {value.shape}")
    print("-" * 20)

print("Metadata:")
for key, value in metadata.items():
    if key == 'feature_chunk':
        # Convert the EagerTensor to a numpy array, then to a string
        decoded_value = value.numpy()[0].decode('utf-8')
        print(f"Key: {key}")
        print(f"Value: {decoded_value}")
    else:
        print(f"Key: {key}")
        print(f"Value: {value.numpy()}")  # Convert tensor to numpy for display
    print("-" * 20)```

### DIFF
--- a/usl_models/usl_models/flood_ml/metastore.py
+++ b/usl_models/usl_models/flood_ml/metastore.py
@@ -151,11 +151,11 @@ def get_spatial_feature_and_label_chunk_metadata(
 def get_temporal_feature_metadata_for_prediction(
     db: firestore.Client, config_name: str
 ) -> dict[str, Any]:
-    """Retrieves metadata stating the location of temporal features in GCS.
+    """Retrieves metadata for temporal features in GCS.
 
     Args:
       db: The firestore client to use when retrieving metadata.
-      config_name: The name of the city cat configuration for which to retrieve metadata.
+      config_name: The name of the configuration for which to retrieve metadata.
 
     Returns:
       A dictionary with keys 'as_vector_gcs_uri' and 'rainfall_duration' which state the
@@ -165,53 +165,44 @@ def get_temporal_feature_metadata_for_prediction(
     Raises:
       ValueError: If a configuration `config_name` cannot be found.
     """
-    collection_name = "city_cat_rainfall_configs"
-    config_ref = db.collection(collection_name).document(config_name)
+    config_ref = db.collection("city_cat_rainfall_configs").document(config_name)
     config = config_ref.get()
 
     if not config.exists:
         raise ValueError(f"No such config {config_name} found.")
 
-    config_data = config.to_dict()
-   
-    return config_data
+    return config.to_dict()
 
 
 def get_spatial_feature_chunk_metadata_for_prediction(
     db: firestore.Client, study_area: str
 ) -> list[dict[str, Any]]:
-    """Retrieves metadata for chunks in the NYC study area.
+    """Retrieves metadata for chunks in the given study area.
 
     Args:
       db: The firestore client to use when retrieving metadata.
-      study_area: The name of the study_area, ex: NYC
+      study_area: The name of the study_area, e.g. NYC
 
     Returns:
-      A list of dictionaries, each representing a chunk with its metadata.
+      A list of dictionaries, each representing a chunk's metadata.
 
     Raises:
-      ValueError: If the NYC document or its chunks collection cannot be found.
+      ValueError: If the study area can not be found or it has no chunks.
     """
-    # Get the reference to the NYC document in the study_area collection
     doc_ref = db.collection("study_areas").document(study_area)
     doc = doc_ref.get()
 
     if not doc.exists:
-        raise ValueError("NYC document not found in study_area collection.")
+        raise ValueError(f"No such study area '{study_area}' found.")
 
-    # Get the chunks collection under the NYC document
-    chunks_collection = doc_ref.collection("chunks")
-    
-    # Retrieve all chunks
-    chunks = chunks_collection.stream()
-    
-    # Convert each chunk document to a dictionary and store in a list
+    chunks = doc_ref.collection("chunks").stream()
     chunk_metadata = [chunk.to_dict() for chunk in chunks]
-    
+
     if not chunk_metadata:
-        raise ValueError(f"No chunks found in the {doc_ref} document.")
+        raise ValueError(f"No chunks found in study area '{study_area}'.")
 
     return chunk_metadata
+
 
 _T = TypeVar("_T")
 
--- a/usl_models/usl_models/flood_ml/metastore.py
+++ b/usl_models/usl_models/flood_ml/metastore.py
@@ -148,6 +148,71 @@ def get_spatial_feature_and_label_chunk_metadata(
     ]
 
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: usl_models/usl_models/flood_ml/metastore.py [hunk 1]
```diff
@@ -107,6 +107,71 @@ def get_spatial_feature_and_label_chunk_metadata(
...
+
+    Raises:
+      ValueError: If the NYC document or its chunks collection cannot be found.
+    """
+    # Get the reference to the NYC document in the study_area collection
+    doc_ref = db.collection("study_areas").document(study_area)
+    doc = doc_ref.get()
+
+    if not doc.exists:
+        raise ValueError("NYC document not found in study_area collection.")
```
waltaskew, line 159: We should get the study area name in here -- something like
```python
raise ValueError(f"No such study area {study_area} found.")
```

usl_models/usl_models/flood_ml/metastore.py
```diff
@@ -126,4 +191,4 @@ def _get_simulation_doc(
             urllib.parse.unquote(sim_name),
             safe=(),
         )
-    )
+    )
```
waltaskew, line 194: You should configure your editor to place newlines at the end of files, as decreed by the laws of UNIX.

usl_models/usl_models/flood_ml/prediction_dataset.py
```diff
@@ -0,0 +1,195 @@
...
+      firestore_client: The client to use when interacting with Firestore.
+      storage_client: The client to use when interacting with Cloud Storage.
+    """
+    firestore_client = firestore_client or firestore.Client()
+    storage_client = storage_client or storage.Client()
+
+    def generator():
+        """Generator for producing full inputs from study area"""
+        for model_input, metadata in _iter_model_inputs_for_prediction(
+        firestore_client,
```
waltaskew, line 54: These should be indented. Running [black](https://github.com/psf/black) should take care of this for you.

usl_models/usl_models/flood_ml/prediction_dataset.py
```diff
@@ -0,0 +1,195 @@
...
+            ),
+          )
+    # If no batch specified, do not batch the dataset, which is required
+    # for generating data for batch prediction in VertexAI.
+    if batch_size:
+        print("batch: ", batch_size)
+        dataset = dataset.batch(batch_size)
+    return dataset
+
+def _generate_temporal_tensor(
```
waltaskew, line 104: A lot of this code looks copy-pasted from dataset.py
We should refactor the common parts of the code used for building prediction & training datasets. It looks like the primary difference is whether we're calling

get_spatial_feature_chunk_metadata_for_prediction
vs
get_spatial_feature_chunk_metadata

or
 
get_temporal_feature_metadata_for_prediction
vs
get_temporal_feature_metadata

and the code should be otherwise nearly identical.

Katsutoshii: Perhaps a better API is to pass a bool arg `training=False` for prediction case, which is pretty common in Keras libs.

Katsutoshii:
@prashantkul can you fix the USL Models CI checks?

usl_models/usl_models/flood_ml/metastore.py
prashantkul: Agreed.

usl_models/usl_models/flood_ml/metastore.py
prashantkul: Will do.

usl_models/usl_models/flood_ml/prediction_dataset.py
prashantkul: I have black running but maybe i need to run  `Format` specifically but good to know.

usl_models/usl_models/flood_ml/prediction_dataset.py
prashantkul: I will look into refactoring this.

prashantkul:
> @prashantkul can you fix the USL Models CI checks?

Not until Thursday.

usl_models/usl_models/flood_ml/prediction_dataset.py
prashantkul: @waltaskew how would you like to me refactor this? The prediction functions takes `city_cat_config` name for retrieving temporal tensors and the `study_area` for feature tensors.

waltaskew: It looks like there is a fair amount of copy-pasted code between the two. A small example -- both have copies of `_download_as_array` and `_download_as_tensor`. The two files should be importing and calling the same function rather than having duplicate copies of the function.
Generally, the more duplicated logic between the two files can be factored into shared functions used by both, the better.

prashantkul: Ah, yes, i can change that for sure - my bad on that one. Is there any suggestion on the main functions?

waltaskew:
It looks like @rsutormin is implementing similar logic over here:
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/150

Let's make sure we're not duplicating further effort.

prashantkul:
> It looks like @rsutormin is implementing similar logic over here: #150
> 
> Let's make sure we're not duplicating further effort.

Looks like @rsutormin also coded this class at a later time? They both do the same thing. My changes are not needed if his changes look good. I can delete my branch. @rsutormin can you confirm?

waltaskew:
> > It looks like @rsutormin is implementing similar logic over here: #150
> > Let's make sure we're not duplicating further effort.
> 
> Looks like @rsutormin also coded this class at a later time? They both do the same thing. My changes are not needed if his changes look good. I can delete my branch. @rsutormin can you confirm?

I think the code we want is actually a mix of the two. You've got the study_area & city_cat_config name interface for load_dataset that we want and you added some code for looking up the metadata associated with them. That's logic needed in @rsutormin 's code.

usl_models/usl_models/flood_ml/prediction_dataset.py
prashantkul: I figured out the issue with black. I switched VM to use tf2.14 and that didn't install black in VSCode's remote SSH session. I was wondering why I was not harasses by it ;)

waltaskew:
@Katsutoshii I implemented the feedback I'd given on my review, added some tests, got the build passing and refactored some duplicated logic. 
I think I'd actually prefer to merged dataset.py & prediction_dataset.py into one file, but for now I'm keeping the same interface you're using.

Katsutoshii (approved):
LGTM

PR #149, opened 2024-07-23
[WRF] Create CF to write heat config metadata

saskatchuwan:
Process flow:
1. WRF team will manually create `Heat_Data_<Sim_Data_Year>.txt` file with metadata (sim year, percentile, etc)
2. WRF team will upload heat data files to `climateiq-atmospheric-simulation-config` with directory structure:
```
/<study_area_name>
    /<config_group_name>
        <file>.txt

Example:
/NYC_Heat
    /Summer_Config_Group (config group name can be whatever modeling team wants)
       Heat_Data_2012.txt
```
4. On file upload, new CF `write_heat_scenario_config_metadata` will be triggered
5. CF will:
    * Open heat data file, read key/value pairs within data file
    * Populate and write `HeatScenarioConfig` entry to metastore

Note: This CF will do nothing else - it will not kick off sims or write new files into the pipeline. It will simply write the config data to metastore so that data like `percentile` is available later in the pipeline

Later in the pipeline:
1. WRF simulation data are uploaded to `climateiq-atmospheric-simulation-output`
2. CF function `build_wrf_label_matrix` from https://github.com/UrbanSystemsLab/climateiq-cnn/pull/120 will trigger
3. `Simulation` entry will be populated with a path to `HeatScenarioConfig` will be written to metastore


MadhaviJain90 (approved):
lgtm!!

usl_pipeline/usl_lib/usl_lib/readers/config_readers.py
```diff
@@ -21,3 +21,22 @@ def read_rainfall_amounts(rain_fd: TextIO) -> Sequence[float]:
             continue
         entries.append(float(rainfall_line.group(1)))
     return entries
+
+
+def read_key_value_pairs(txt_fd: TextIO) -> dict:
```
waltaskew, line 26: If we're creating the config file and deciding its format, I think we should pick an existing format with existing readers & writers rather than build our own?

What you've got looks pretty similar to an ini file
https://docs.python.org/3/library/configparser.html

I belive you can create an ini file without the [section] parts, which looks like exactly what you're supporting here.

so I'd just go with that existing ini support, or something like JSON or YAML

usl_pipeline/usl_lib/tests/storage/test_metastore.py
```diff
@@ -242,6 +242,47 @@ def test_flood_scenario_config_delete():
     )
 
 
+def test_heat_scenario_config_set():
```
waltaskew, line 245: I'd generally advocate for writing integration tests against a local firestore emulator for more realistic tests, such as you see in here:
https://github.com/UrbanSystemsLab/climateiq-cnn/blob/main/usl_pipeline/usl_lib/tests/integration/test_metastore_integration.py#L34

usl_pipeline/usl_lib/usl_lib/readers/config_readers.py
saskatchuwan: Gotcha - so the contents can stay the same but use `.ini` file extension and replace the custom reader with the configparser lib? Or could we user configparser on .txt files as well?

waltaskew: the parser won't care if the extension of the file is .ini or .txt

usl_pipeline/usl_lib/usl_lib/readers/config_readers.py
```diff
@@ -21,3 +23,40 @@ def read_rainfall_amounts(rain_fd: TextIO) -> Sequence[float]:
...
+
+    config_dict = {}
+
+    # Iterate through all sections and add key-value pairs directly to config_dict
+    for section in config.sections():
+        for key, value in config.items(section):
+            config_dict[key] = value
+
+    # If there are no sections, add key-value pairs from the DEFAULT section
+    if not config.sections():
```
waltaskew, line 58: How would this happen? The '[DEFAULT]' section we added should be picked up in the above loop. I don't think we need this branch? I think we can get away with just

```python
config = configparser.ConfigParser()
config.read_file(itertools.chain(['[DEFAULT]\n'], fd))
return dict(itertools.chain(config.items(section) for section in config.section()))
```
or even
```python
config = configparser.ConfigParser()
config.read_file(itertools.chain(['[DEFAULT]\n'], fd))
return config['DEFAULT']
```
if we know there won't be any sections in the file we're processing since the `config['DEFAULT']` will return a dict-like object already

usl_pipeline/usl_lib/usl_lib/readers/config_readers.py
```diff
@@ -21,3 +23,40 @@ def read_rainfall_amounts(rain_fd: TextIO) -> Sequence[float]:
...
+      fd: The file containing the key/value pairs
+
+    Returns:
+      A dictionary of the key/values pairs in the file
+    """
+    # Read lines from file first. If the first line is not blank and doesn't look
+    # like a section header, add [DEAULT] section so it is compliant with
+    # what ConfigParser expects
+    lines = fd.readlines()
+    if lines and not lines[0].strip().startswith("["):
```
waltaskew, line 41: You don't need to do this conditionally -- it'd just be adding an empty DEFAULT section if there is already a section in the ini file.

You can do this in one line with:

```python
config.read_file(itertools.chain(['[DEFAULT]\n'], fd))
```

usl_pipeline/cloud_functions/main.py
```diff
@@ -220,6 +220,42 @@ def _get_crs_from_wps(nc_dataset: netCDF4.Dataset) -> str:
...
+
+        metastore.HeatScenarioConfig(
+            name=config_blob.name,
+            # File names should be in the form:
+            # <study_area>/<parent_config_name>/<file_name>
+            # Ex: NYC_Heat/Summer_Config/Heat_Data_2012.txt
+            parent_config_name=file_name.parent.name,
+            gcs_uri=f"gs://{config_blob.bucket.name}/{config_blob.name}",
+            simulation_year=int(config_dict["simulation_year"]),
+            simulation_months=config_dict["simulation_months"],
```
waltaskew, line 254: what is 'simulation_months' ? is it not an integer as well?

usl_pipeline/cloud_functions/main.py
```diff
@@ -220,6 +220,42 @@ def _get_crs_from_wps(nc_dataset: netCDF4.Dataset) -> str:
...
+            percentile=int(config_dict["percentile"]),
```
waltaskew, line 255: is the file format for this documented? folks might expect this to be a float `.99` rather than an int `99` to express 99th percentile. I don't feel strongly either way -- just want to make sure it's written down because it is ambiguous.

usl_pipeline/cloud_functions/requirements.txt
```diff
@@ -9,6 +9,7 @@ click==8.1.7
 click-plugins==1.1.1
 cligj==0.7.2
 cloudevents==1.10.1
+configparser==7.0.0
```
waltaskew, line 12: configparser is in the stdlib. We don't need this requirements entry.
It looks like that package is to provide the newest version of configparser to older versions of python.

usl_pipeline/cloud_functions/main.py
saskatchuwan: nope! preference is to use a convention they have in climate science (JJA = June July August)

usl_pipeline/cloud_functions/main.py
saskatchuwan: not yet! working on one that will provide example of file format

MadhaviJain90 (approved):
LGTM!!

usl_pipeline/usl_lib/usl_lib/readers/config_readers.py
saskatchuwan: thanks - that simplifies it. was originally trying to support regular ini files with headers but i don't think we'll need it. we will just assume the files do not include the headers

waltaskew (approved):
LGTM

usl_pipeline/usl_lib/usl_lib/readers/config_readers.py
```diff
@@ -21,3 +23,26 @@ def read_rainfall_amounts(rain_fd: TextIO) -> Sequence[float]:
...
+
+    Returns:
+      A dictionary of the key/values pairs in the file
+    """
+    # Read lines from file first. If the first line is not blank and doesn't look
+    # like a section header, add [DEAULT] section so it is compliant with
+    # what ConfigParser expects
+    config = configparser.ConfigParser()
+    config.read_file(itertools.chain(["[DEFAULT]\n"], fd))
+    config_dict = {key: value for key, value in config["DEFAULT"].items()}
```
waltaskew, line 47: It'd also be fine to `return config["DEFAULT"]` because `config["DEFAULT"]` is a dict-like (e.g. you can do `config["DEFAULT"]['simulation_year']`

can use the more generic Mapping
```python
read_key_value_pairs(fd: TextIO) -> Mapping
```
from the typing module to indicate a dict-like rather than specifically a dict

waltaskew: but what you've got here is fine!

PR #150, opened 2024-07-23
Dataset for predictions

rsutormin:


usl_models/usl_models/flood_ml/prediction_dataset.py
```diff
@@ -0,0 +1,107 @@
...
+from usl_models.flood_ml import constants, dataset
+
+
+class PredictionInput(TypedDict):
+    """Input tensors dictionary used for prediction."""
+
+    geospatial: tf.Tensor
+    temporal: tf.Tensor
+    spatiotemporal: tf.Tensor
+    metadata: tf.Tensor
```
waltaskew, line 19: I'd just add metadata to the existing `FlodModel.Input` dict. Maybe make it an Optional field. If training ignores the metadata we supply, that's fine. Doesn't hurt to add it since we need to look up the metadata anyway.

I think that would help share more code between the two dataset loading code paths.

usl_models/usl_models/flood_ml/prediction_dataset.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def load_prediction_dataset(
+    rainfall_config_gcs_uri: str,
+    chunk_id_to_gcs_uri_map: Dict[str, str],
```
waltaskew: I think the interface we want is what @prashantkul  implemented over here:

https://github.com/UrbanSystemsLab/climateiq-cnn/pull/145/files#diff-d2183c123f3657caeb8ad0386b9f4e6aad151f39f95fcc0c2dbaf8d22678d705R21

where the caller provides the study area and config name, and the function itself is responsible for doing metastore lookups.

waltaskew:
obviated by https://github.com/UrbanSystemsLab/climateiq-cnn/pull/145

PR #152, opened 2024-07-25
[WRF] Update wind derivation to split out cos/sin components

saskatchuwan:
Update our wind direction computation to split out sin and cos components - code extracted from source: https://github.com/UrbanSystemsLab/climateiq-cnn/pull/114

### DIFF
--- a/usl_pipeline/cloud_functions/main_test.py
+++ b/usl_pipeline/cloud_functions/main_test.py
@@ -603,14 +603,14 @@ def test_process_wps_feature_apply_minmax_scaler():
     )
 
 
-def test_compute_custom_wps_variables_wind():
+def test_compute_wind_components():
     ncfile = netCDF4.Dataset("met_em_test.nc", mode="w", format="NETCDF4", memory=1)
     ncfile.createDimension("Time", 1)
-    ncfile.createDimension("west_east", 3)
-    ncfile.createDimension("south_north", 3)
-    ncfile.createDimension("west_east_stag", 4)
-    ncfile.createDimension("south_north_stag", 4)
-    ncfile.createDimension("num_metgrid_levels", 3)
+    ncfile.createDimension("west_east", 2)
+    ncfile.createDimension("south_north", 2)
+    ncfile.createDimension("west_east_stag", 3)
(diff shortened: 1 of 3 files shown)


usl_pipeline/cloud_functions/main_test.py
```diff
@@ -636,9 +636,9 @@ def test_compute_custom_wps_variables_wind():
     processed_ds = main._compute_custom_wps_variables(ds)
 
     # Check that processed dataset contains newly computed variables
-    assert all(var in processed_ds.keys() for var in ["WSPD10", "WDIR10"])
+    assert all(var in processed_ds.keys() for var in ["WSPD", "WDIR_SIN", "WDIR_COS"])
```
waltaskew, line 639: should we have some tests for the computed values as well?

waltaskew (approved):
hooray for trigonometry!

usl_pipeline/cloud_functions/main_test.py
saskatchuwan: @MadhaviJain90 is this something you can help validate?

waltaskew: can you use some simpler wind vectors as test inputs? like, one that runs exactly east?

MadhaviJain90: The only thing that is concerning to me is that all values of sin and cos are always negative? The scaling was from 1 to -1. So we should be seeing some positive values too

waltaskew: although not knowing exactly what they represent, I'd assume you'll get simpler values to reason about with unit vectors like [1, 0, 0] & such

saskatchuwan: Thanks for the suggestion. updated the test. feels right but can't confirm my trig is bad
@MadhaviJain90 take a look at the expected values in test assertions. if it looks good, i'll merge!

MadhaviJain90 (approved):
All good 👍

usl_pipeline/cloud_functions/main_test.py
saskatchuwan: Perf!

PR #153, opened 2024-07-25
Support loading a model from an artifact URI.

Katsutoshii:
Support loading a model from an artifact URI.

### DIFF
--- a/usl_models/requirements.txt
+++ b/usl_models/requirements.txt
@@ -8,7 +8,7 @@ gast==0.5.4
 google-api-core==2.19.1
 google-auth==2.32.0
 google-auth-oauthlib==1.0.0
-google-cloud-aiplatform==1.43.0
+google-cloud-aiplatform[prediction]==1.43.0
 google-cloud-bigquery==3.25.0
 google-cloud-core==2.4.1
 google-cloud-firestore==2.15.0
--- a/usl_models/requirements.txt
+++ b/usl_models/requirements.txt
@@ -8,7 +8,7 @@ gast==0.5.4
 google-api-core==2.19.1
 google-auth==2.32.0
 google-auth-oauthlib==1.0.0
-google-cloud-aiplatform[prediction]==1.43.0
+google-cloud-aiplatform==1.43.0
 google-cloud-bigquery==3.25.0
 google-cloud-core==2.4.1
 google-cloud-firestore==2.15.0
--- a/usl_models/usl_models/flood_ml/model.py
+++ b/usl_models/usl_models/flood_ml/model.py
@@ -1,11 +1,11 @@
 """Flood model definition."""
 
 import logging
+import os
 from typing import TypeAlias, TypedDict, List, Callable
 
 import keras
 from keras import layers
-from google.cloud.aiplatform.utils import prediction_utils
 import tensorflow as tf
 
 from usl_models.flood_ml import constants
@@ -32,24 +32,40 @@ class FloodModel:
 
     def __init__(
         self,
-        model_params: FloodModelParams,
+        params: FloodModelParams | None,
         spatial_dims: tuple[int, int] = (constants.MAP_HEIGHT, constants.MAP_WIDTH),
-        artifact_uri: str | None = None,
     ):
         """Creates the flood model.
 
         Args:
-            model_params: A dictionary of configurable model parameters.
+            params: A dictionary of configurable model parameters.
             spatial_dims: Tuple of spatial height and width input dimensions.
                 Needed for defining input shapes. This is an optional arg that
                 can be changed (primarily for testing/debugging).
             artifact_uri: Optional artifact URI to load model weights from.
         """
-        self._model_params = model_params
+        self._model_params = params or model_params.default_params()
         self._spatial_dims = spatial_dims
         self._model = self._build_model()
-        if artifact_uri is not None:
-            self._load_weights_from_artifact(artifact_uri)
+
+    @classmethod
+    def from_checkpoint(cls, artifact_uri: str, **kwargs) -> "FloodModel":
+        """Loads the model from a checkpoint URI.
+
+        We load weights only to keep custom methods (e.g. `call_n`) intact.
+        This only works if the model architecture is identical to the architecure
+        used during export.
+        Ideally, we would load the entire Keras model and use that directly to allow
+        loading different architectures within the same wrapper class.
+        Unfortunately, `call_n` is not trivially serializeable in its current state.
+        """
+        model = cls(**kwargs)
+        loaded_model = keras.models.load_model(os.path.join(artifact_uri, "model"))
+        if loaded_model is not None:
+            model._model.set_weights(loaded_model.get_weights())
+        else:
+            logging.error(f"Failed to load model from: {artifact_uri}")
+        return model
 
     def _build_model(self) -> keras.Model:
         """Creates the correct internal (Keras) model architecture."""
@@ -66,12 +82,6 @@ class FloodModel:
             ],
         )
         return model
-
-    def _load_weights_from_artifact(self, artifact_uri: str):
-        """Loads model weights from a GCS artifact URI."""
-        prediction_utils.download_model_artifacts(artifact_uri)
-        loaded_model = keras.models.load_model("model")
-        self._model.set_weights(loaded_model.get_weights())
 
     def call(self, input: Input) -> tf.Tensor:
         """Predict the next timestep. See `FloodConvLSTM.call`."""
@@ -341,6 +351,7 @@ class FloodConvLSTM(tf.keras.Model):
 
         # We use 1-indexing for simplicity. Time step t represents the t-th flood
         # prediction.
+        # TODO: consider using tf.while_loop to support serializing this function.
         for t in range(1, n + 1):
             input = FloodModel.Input(
                 geospatial=geospatial,
--- a/usl_models/usl_models/flood_ml/model.py
+++ b/usl_models/usl_models/flood_ml/model.py
@@ -31,7 +31,7 @@ class FloodModel:
 
     def __init__(
         self,
-        params: FloodModelParams | None,
+        params: FloodModelParams | None = None,
         spatial_dims: tuple[int, int] = (constants.MAP_HEIGHT, constants.MAP_WIDTH),
     ):
         """Creates the flood model.
--- a/usl_models/usl_models/flood_ml/model.py
+++ b/usl_models/usl_models/flood_ml/model.py
@@ -1,7 +1,6 @@
 """Flood model definition."""
 
 import logging
-import os
 from typing import TypeAlias, TypedDict, List, Callable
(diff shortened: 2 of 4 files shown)


usl_models/requirements.txt
```diff
@@ -8,7 +8,7 @@ gast==0.5.4
 google-api-core==2.19.1
 google-auth==2.32.0
 google-auth-oauthlib==1.0.0
-google-cloud-aiplatform==1.43.0
+google-cloud-aiplatform[prediction]==1.43.0
```
waltaskew, line 11: What additional packages does the `[prediction`] bit pull in? They should make it into this requirements.txt as well. We'll also need to update the setup.py, as that's what's used when we deploy this code into training land, not the requirements.txt (we should maybe remove the requirements.txt as it's not used to deploy anything at the moment) to make clear the `setup.py` is the important bit.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -39,6 +34,7 @@ def __init__(
         self,
         model_params: FloodModelParams,
         spatial_dims: tuple[int, int] = (constants.MAP_HEIGHT, constants.MAP_WIDTH),
```
waltaskew, line 36: If we're loading from a model check point, what does the user pass for model_params & spatial_dims? What happens if they're different from the params & dims used to train that model checkpoint?

I'm wondering if the FloodModel class is what we should be working with rather than working with the underlying keras model itself.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -68,57 +67,11 @@ def _build_model(self) -> keras.Model:
...
-            st_shape = data.geospatial.shape[:3] + [1]
-            st_input = tf.zeros(st_shape)
-
-        data = dataclasses.replace(data, spatiotemporal=st_input)
-        return data
+    def _load_weights_from_artifact(self, artifact_uri: str):
+        """Loads model weights from a GCS artifact URI."""
+        prediction_utils.download_model_artifacts(artifact_uri)
+        loaded_model = keras.models.load_model("model")
+        self._model.set_weights(loaded_model.get_weights())
```
waltaskew, line 74: Is it just the weights we need? Or do we need `self._model = loaded_model` ?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -68,57 +67,11 @@ def _build_model(self) -> keras.Model:
...
-        if st_input is None:
-            st_shape = data.geospatial.shape[:3] + [1]
-            st_input = tf.zeros(st_shape)
-
-        data = dataclasses.replace(data, spatiotemporal=st_input)
-        return data
+    def _load_weights_from_artifact(self, artifact_uri: str):
+        """Loads model weights from a GCS artifact URI."""
+        prediction_utils.download_model_artifacts(artifact_uri)
+        loaded_model = keras.models.load_model("model")
```
waltaskew, line 73: We can do 
```python
 keras.models.load_model('gs://climateiq-vertexai/aiplatform...')
```
to grab the model without using `download_model_artifacts`
I'd prefer not to use `download_model_artifacts` because it writes things out to the local disk.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -39,6 +34,7 @@ def __init__(
         self,
         model_params: FloodModelParams,
         spatial_dims: tuple[int, int] = (constants.MAP_HEIGHT, constants.MAP_WIDTH),
+        artifact_uri: str | None = None,
```
waltaskew, line 37: Might suggest something like
```python
def __init__(self, model=None):
  self._model = model or self._build_model()

@classmethod
def from_checkpoint(cls, artifact_uri):
  model = keras.models.load_model(artifact_uri)
  return cls(model=model)
```

usl_models/requirements.txt
Katsutoshii: This is required for prediction_util which lets us download the model artifacts from artifact registry. I'll move this to setup.py.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: The model params need to match what was used during export, otherwise it will throw an error referring to the shape of the weights.
We can't work with the loaded Keras model directly because we were unable to export its extra methods (call_n) and the signatures get mangled upon export (the dict becomes a kwargs mapping).

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Loading the weights is the only way to keep the custom keras.Model class intact due to the aforementioned changes upon export.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Gotcha, I didn't realize this was possible. Good to know!

usl_models/usl_models/flood_ml/model.py
Katsutoshii: How would this API support changing the parameters?

usl_models/usl_models/flood_ml/model.py
waltaskew: so it sounds like we'd need to persist the `model_params` and `spatial_dims` as well to be able to reliably load the model? Or else get the call_n signature working -- do you see any obvious issues with what I'm doing wrong here:
https://github.com/UrbanSystemsLab/climateiq-cnn/pull/127
where I attempt to add that signature?

usl_models/usl_models/flood_ml/model.py
waltaskew: you can add model parameters to the above methods -- the crux of what I'm suggesting is having an explicit
```
@classmethod
def from_checkpoint
```
method for loading a model from GCS rather than overloading the constructor with additional behaviours

usl_models/usl_models/flood_ml/model.py
waltaskew: okay, some comments describing these issues would be helpful

this means that if the model architecture changes, old models essentially become un-loadable, right? Because they'd no longer be compatible with what `self._build_model()` produces?

Katsutoshii: Sure, will document this.

waltaskew: Am I correct that this makes old models un-loaded-able if the architecture changes? That seems like a problem for experimentation.
Is it just the call_n signature thing we need to fix to make this work?

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Even if we had call_n serialization working, we still need to know the model_params and spatial_dims to provide correctly shaped input data. So the value of making this work for arbitrary model architecture is limitted to convenience of testing different hidden layers.
Testing changes to hidden layers is still easy enough when kicking off a custom training job, the only part that requires some thought is running test predictions, where the user must ensure that they are using the correct branch of the code so the model architecture lines up. 

We weren't able to get call_n to serialize without making major changes to the model code. I think we would need to get rid of the for loop and replace with tf.while_loop so the resulting tf.Graph doesn't unroll the loop for a fixed n during serialization. This will take some work and testing, so given the time constraint it's easiest to leave that out of serialization.

waltaskew: okay, that sounds good -- let's just focus on documenting all the issues you've laid out here

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Under this setup, models will need to be loaded from the same version of the code they were saved in.
We could try moving call_n logic into the wrapper class. Then the only remaining issue is unwrapping the inputs dictionary into kwargs to match what is expected from serving_default.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
waltaskew: Okay, let's just continue focusing on documenting issues rather than solving issues. As long as we make clear what issues are and suggest some path forward for addressing them, I think we're leaving things in a good place.

usl_models/requirements.txt
Katsutoshii: Ack, removed

usl_models/usl_models/flood_ml/model.py
```diff
@@ -37,21 +32,41 @@ class Input(TypedDict):
...
+        Ideally, we would load the entire Keras model and use that directly to allow
+        loading different architectures within the same wrapper class.
+        Unfortunately, `call_n` is not trivially serializeable in its current state.
+        """
+        model = cls(**kwargs)
+        loaded_model = keras.models.load_model(os.path.join(artifact_uri, "model"))
+        if loaded_model is not None:
+            model._model.set_weights(loaded_model.get_weights())
+        else:
+            logging.error(f"Failed to load model from: {artifact_uri}")
```
waltaskew, line 67: should we raise an exception here?
this looks like a failure we can't really recover from

usl_models/usl_models/flood_ml/model.py
```diff
@@ -37,21 +32,41 @@ class Input(TypedDict):
...
+
+        We load weights only to keep custom methods (e.g. `call_n`) intact.
+        This only works if the model architecture is identical to the architecure
+        used during export.
+        Ideally, we would load the entire Keras model and use that directly to allow
+        loading different architectures within the same wrapper class.
+        Unfortunately, `call_n` is not trivially serializeable in its current state.
+        """
+        model = cls(**kwargs)
+        loaded_model = keras.models.load_model(os.path.join(artifact_uri, "model"))
```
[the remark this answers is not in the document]
Katsutoshii: Turns out neither of these approaches work. PurePosixPath will collapse `//` -> `/`, and urljoin doesn't recognize `gs://` as a valid URL prefix.
Let's just expect `/model` in the input URI.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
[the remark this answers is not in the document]
waltaskew: is that because `artificat_uri` has a trailing `/` ? Can do `artifact_uri.rstrip('/')`

waltaskew (approved):

usl_models/usl_models/flood_ml/model.py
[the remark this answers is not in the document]
Katsutoshii: `rstrip` would work. But I think it's better for this to be handled at the callsite in case other file formats are supported later or if the model export logic changes.

PR #154, opened 2024-07-26
Implement batch prediction logic for FloodModel

Katsutoshii:
Implement batch prediction logic for FloodModel

### DIFF
--- a/usl_models/tests/flood_ml/mock_dataset.py
+++ b/usl_models/tests/flood_ml/mock_dataset.py
@@ -107,7 +107,8 @@ def mock_prediction_dataset(
         height: spatial map height
         width: spatial map width
         batch_count: number of batches to produce
-        batch_size: size of each batch. If None, the data is not batched (no batch dimension).
+        batch_size: size of each batch. If None, the data is not batched
+            (no batch dimension).
         n: optional number of timesteps.
             Required for producing data for use with `model.call_n`.
 
--- a/usl_models/tests/flood_ml/mock_dataset.py
+++ b/usl_models/tests/flood_ml/mock_dataset.py
@@ -11,11 +11,13 @@ def input_signature(
     params: model_params.FloodModelParams,
     height: int = 100,
     width: int = 100,
-    n: int = 0,
+    n: int | None = None,
 ) -> dict[str, tf.TensorSpec]:
     """Returns the input signature for a dataset with n timesteps."""
     # If the datset specifies a number of timesteps, temporal tensor is of max size.
-    temporal_duration = constants.MAX_RAINFALL_DURATION if n else params["n_flood_maps"]
+    temporal_duration = (
+        constants.MAX_RAINFALL_DURATION if n is not None else params["n_flood_maps"]
+    )
     return dict(
         geospatial=tf.TensorSpec(
             shape=(height, width, constants.GEO_FEATURES), dtype=tf.float32
@@ -29,9 +31,11 @@ def input_signature(
     )
 
 
-def label_signature(height: int = 100, width: int = 100, n: int = 0) -> tf.TensorSpec:
+def label_signature(
+    height: int = 100, width: int = 100, n: int | None = None
+) -> tf.TensorSpec:
     """Returns the label signature for a dataset with n timesteps."""
-    shape = (n, height, width) if n else (height, width)
+    shape = (n, height, width) if n is not None else (height, width)
     return tf.TensorSpec(shape=shape, dtype=tf.float32)
 
 
@@ -48,8 +52,8 @@ def mock_dataset(
     height: int = 100,
     width: int = 100,
     batch_count: int = 1,
-    batch_size: int = 1,
-    n: int = 0,
+    batch_size: int | None = 1,
+    n: int | None = None,
 ) -> tf.data.Dataset:
     """Constructs a dataset of random mock data.
 
@@ -70,7 +74,8 @@ def mock_dataset(
 
     def generator():
         """Generate random tensors."""
-        for i in range(batch_size * batch_count):
+        batch_size_int = 1 if batch_size is None else batch_size
+        for i in range(batch_size_int * batch_count):
             yield flood_model.FloodModel.Input(
                 geospatial=tf.random.normal(shape=input_sig["geospatial"].shape),
                 temporal=tf.random.normal(shape=input_sig["temporal"].shape),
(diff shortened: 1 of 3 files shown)


usl_models/tests/flood_ml/mock_dataset.py
```diff
@@ -77,3 +85,47 @@ def generator():
...
     return dataset
+
+
+def mock_prediction_dataset(
+    params: model_params.FloodModelParams,
+    height: int = 100,
+    width: int = 100,
+    batch_count: int = 1,
+    batch_size: int = 1,
+    n: int = 0,
```
tatianawu, line 96: Why is the default zero? The docstring says this is optional, so should we default to None instead?

usl_models/tests/flood_ml/mock_dataset.py
```diff
@@ -77,3 +85,47 @@ def generator():
...
+                temporal=tf.random.normal(shape=input_sig["temporal"].shape),
+                spatiotemporal=tf.random.normal(
+                    shape=input_sig["spatiotemporal"].shape
+                ),
+            ), dict(feature_chunk=str(i), rainfall=i)
+
+    dataset = tf.data.Dataset.from_generator(
+        generator=generator, output_signature=(input_sig, metadata_sig)
+    )
+    if batch_size:
```
tatianawu, line 129: Won't this always be true, unless we set batch size 0? It seems strange to take 0 if we don't want to batch -- should this be optional instead?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -128,6 +134,62 @@ def call_n(self, full_input: Input, n: int = 1) -> tf.Tensor:
...
+                batch_predictions = []
+                for batches in replica_batches:
+                    for batch in batches:
+                        batch_predictions.append(batch)
+
+            results = []
+            for prediction, chunk_id in zip(
+                batch_predictions, metadata["feature_chunk"]
+            ):
+                # print("metadata:", metadata)
```
tatianawu, line 185: Looks like there are some debugging print statements. Should we clean these up?

If we want to print anything in the logs, I think doing that via `logging` would be better.

usl_models/tests/flood_ml/mock_dataset.py
Katsutoshii: Sure, we can use None instead.

usl_models/tests/flood_ml/mock_dataset.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: This was left over from testing, removed.

usl_models/tests/flood_ml/mock_dataset.py
```diff
@@ -11,11 +11,13 @@ def input_signature(
     params: model_params.FloodModelParams,
     height: int = 100,
     width: int = 100,
-    n: int = 0,
+    n: int | None = None,
```
waltaskew, line 14: I think it'd be helpful to use a more descriptive name than 'n' here. Here it looks like the rainfall duration, while the 'n' below in label_signature looks like the batch size, and the 'n' in mock_dataset is rainfall duration.

Also, I think the rest of our code is doing `Optional[X]` rather than `X | None` so would like to stay consistent

usl_models/tests/flood_ml/mock_dataset.py
```diff
@@ -77,3 +90,48 @@ def generator():
...
+                temporal=tf.random.normal(shape=input_sig["temporal"].shape),
+                spatiotemporal=tf.random.normal(
+                    shape=input_sig["spatiotemporal"].shape
+                ),
+            ), dict(feature_chunk=str(i), rainfall=i)
+
+    dataset = tf.data.Dataset.from_generator(
+        generator=generator, output_signature=(input_sig, metadata_sig)
+    )
+    if batch_size is not None:
```
waltaskew, line 135: it looks like this leads to a different behaviour if batch_size is 1 or None? Is that intentional? Might be worth documenting if that's intentional. If not, I'd just have `batch_size: int = 1` so that there aren't two ways to express the same thing.

usl_models/usl_models/flood_ml/model.py
```diff
@@ -128,6 +134,60 @@ def call_n(self, full_input: Input, n: int = 1) -> tf.Tensor:
...
+
+        Args:
+            strategy: multi-GPU distribute strategy.
+            dataset: Dataset generating (inputs, metadata) tuples.
+            n: number of timesteps to predict.
+
+        Returns: an iterator containing batches of results.
+        """
+        strategy = tf.distribute.get_strategy()
+        dataset = strategy.experimental_distribute_dataset(dataset)
```
waltaskew, line 162: always exciting to see the word `experimental` in a method name :)
how... 'experimental' is this?

usl_models/usl_models/flood_ml/model.py
```diff
@@ -128,6 +134,60 @@ def call_n(self, full_input: Input, n: int = 1) -> tf.Tensor:
...
+            if strategy.num_replicas_in_sync > 1:
+                replica_batches = batch_predictions.values
+                batch_predictions = []
+                for batches in replica_batches:
+                    for batch in batches:
+                        batch_predictions.append(batch)
+
+            results = []
+            for prediction, chunk_id in zip(
+                batch_predictions, metadata["feature_chunk"]
```
waltaskew, line 182: Can you leave a comment explaining why it's okay to assume that batch_predictions will be ordered the same as the metadata? And if there are cases where that would not be the case?

waltaskew (approved):
Exciting to see this coming together!

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/model.py
Katsutoshii: The docs don't discourage it's usage, seems like it's be stable for the past few TF versions: https://www.tensorflow.org/api_docs/python/tf/distribute/Strategy#experimental_distribute_dataset

usl_models/tests/flood_ml/mock_dataset.py
Katsutoshii: The difference is whether a batch dimension is added. I've noted this in the docstring.

usl_models/tests/flood_ml/mock_dataset.py
Katsutoshii: In this file `n` always represents the rainfall duration, which should be clear from docstrings. It's also consistent with the `n` from `call_n`.

We already have `X | None` in the codebase and this is the "modern" 3.10 approach so I prefer to migrate the rest of the codebase at a later time.

PR #155, opened 2024-07-26
Implement BatchPredictor to move remaining logic out of Colab

Katsutoshii:
Implement BatchPredictor to move remaining logic out of Colab


Katsutoshii (commented):
Note that this contains dev implementation of prediction dataset. Will rebase once that has been finalized. In the meantime, please ignore those files.

usl_models/usl_models/flood_ml/batch_predictor.py
```diff
@@ -0,0 +1,191 @@
+"""Utilities for running batch predictions on files stored in GCS."""
+
+import json
+import urllib.parse
+import logging
+import dataclasses
```
waltaskew, line 6: let's alphabetize these stdlib imports

usl_models/usl_models/flood_ml/batch_predictor.py
```diff
@@ -0,0 +1,191 @@
...
+import tensorflow as tf
```
waltaskew, line 7: and group the tf import with the third party imports bellow

usl_models/usl_models/flood_ml/batch_predictor.py
```diff
@@ -0,0 +1,191 @@
...
+    and `run_id`.
+    """
+
+    db: firestore.Client
+    client: storage.Client
+    output_bucket: str
+    model_id: str
+    model: model.FloodModel
+    run_id: str
+    batch_size: int = 1
```
waltaskew, line 31: Are we able to default to a bigger batch size than 1?

usl_models/usl_models/flood_ml/batch_predictor.py
```diff
@@ -0,0 +1,191 @@
...
+            study_area=study_area_id,
+            city_cat_config=parsed_config_id,
+            batch_size=self.batch_size,
+            storage_client=self.client,
+        )
+        chunk_ids = []
+        for results in self.model.batch_predict_n(dataset, n=rainfall_duration):
+            for result in results:
+                chunk_id = result["chunk_id"]
+                if isinstance(chunk_id, tf.Tensor):
```
waltaskew, line 87: when isn't chunk_id a tensor? A comment might be helpful here.

usl_models/usl_models/flood_ml/batch_predictor.py
```diff
@@ -0,0 +1,191 @@
...
+        """Returns a list of chunk metadata."""
+        chunk_metadata = []
+        for chunk_id in chunk_ids:
+            path = self.get_prediction_npy_path(
+                study_area_id=study_area_id, config_id=config_id, chunk_id=chunk_id
+            )
+            chunk_metadata.append(
+                {
+                    "id": chunk_id,
+                    "path": f"gs://{self.output_bucket}/{path}",
```
waltaskew, line 154: let's store the path to the JSONL file we generate which includes these predictions
those .npy files I think we should delete, so they won't be good to link to

usl_models/usl_models/flood_ml/batch_predictor.py
```diff
@@ -0,0 +1,191 @@
...
+        with self.client.bucket(self.output_bucket).blob(jsonl_path).open("w") as fd:
+            for chunk_id in chunk_ids:
+                path = self.get_prediction_npy_path(study_area_id, config_id, chunk_id)
+                logging.info(f"Bundling {path}...")
+                prediction = self.load_prediction_npy(
+                    study_area_id, config_id, chunk_id
+                )
+                json_data = self.prediction_to_json(prediction, chunk_id)
+                json.dump(json_data, fd)
+                fd.write("\n")
```
waltaskew, line 191: Is there a reason to keep the npy files around past this point? Can we delete them now?

waltaskew (commented):
This is great, thanks so much for getting this out to the notebook!
Main overriding comment would be to comment more -- some description on the overall approach (produce intermediate .npy prediction files to ultimately produce a jsonl file as needed by the export pipeline) would be helpful.

usl_models/usl_models/flood_ml/batch_predictor.py
Katsutoshii: Done.

usl_models/usl_models/flood_ml/batch_predictor.py
Katsutoshii: Sure, added a note for how to tune this.

usl_models/usl_models/flood_ml/batch_predictor.py
Katsutoshii: Added a comment.

usl_models/usl_models/flood_ml/batch_predictor.py
Katsutoshii: I've added a path at the model run level.

waltaskew: I think it actually belongs here, because I don't think having one jsonl file for all predictions is an inherent part of the approach (I think we'd actually prefer to generate more smaller files so the export logic can start sooner.)

usl_models/usl_models/flood_ml/batch_predictor.py
Katsutoshii: Sure, I've added delete logic here.

usl_models/usl_models/flood_ml/batch_predictor.py
Katsutoshii: While we are still using a single .jsonl file, do we want each chunk to point to the bundled file?

waltaskew: I think so, because in the future we may (and hopefully will) have multiple jsonl files
I wouldn't want the database schema to overly reflect the current implementation

waltaskew (approved):

PR #156, opened 2024-07-26
Deprecate unused classes and functions.

Katsutoshii:
Deprecate unused classes and functions.
Migrate Optional[X] -> X | None in usl_models.


waltaskew:
I'd really rather not replace `Optional[X]` with `X | None`
`Optional[X]` is the most common usage in the codebase.
I don't want to have different parts of the codebase using different idioms -- I'd rather have usl_models be consistent with our other code.

waltaskew (approved):
Deleted code is the best code!

However, I'd rather not replace `Optional[X]` with `X | None`
`Optional[X]` is the most common usage in the codebase.
I don't want to have different parts of the codebase using different idioms -- I'd rather have usl_models be consistent with our other code.

I have zero opinions on if `Optional[X]` is better or worse than `X | None` -- the only thing I care about is consistency. I'd be fine to change to `X | None` but I'd want to do it in all our code across all our repos.

Katsutoshii:
> Deleted code is the best code!
> 
> However, I'd rather not replace `Optional[X]` with `X | None` `Optional[X]` is the most common usage in the codebase. I don't want to have different parts of the codebase using different idioms -- I'd rather have usl_models be consistent with our other code.
> 
> I have zero opinions on if `Optional[X]` is better or worse than `X | None` -- the only thing I care about is consistency. I'd be fine to change to `X | None` but I'd want to do it in all our code across all our repos.

Makes sense, I've updated to migrate all Optional usage to | None

waltaskew:
> > Deleted code is the best code!
> > However, I'd rather not replace `Optional[X]` with `X | None` `Optional[X]` is the most common usage in the codebase. I don't want to have different parts of the codebase using different idioms -- I'd rather have usl_models be consistent with our other code.
> > I have zero opinions on if `Optional[X]` is better or worse than `X | None` -- the only thing I care about is consistency. I'd be fine to change to `X | None` but I'd want to do it in all our code across all our repos.
> 
> Makes sense, I've updated to migrate all Optional usage to | None

Well, you've actually got a few more repos to go now!
https://github.com/UrbanSystemsLab/climateiq-flood-model
https://github.com/UrbanSystemsLab/climateiq-frontend

PR #158, opened 2024-07-29
Support dropout for training samples with min mean value below a configured threshold.

Katsutoshii:
Support dropout for training samples with min mean value below a configured threshold.

### DIFF
--- a/usl_models/usl_models/flood_ml/dataset.py
+++ b/usl_models/usl_models/flood_ml/dataset.py
@@ -1,5 +1,6 @@
 """tf.data.Datasets for training FloodML model on CityCAT data."""
 
+import dataclasses
 import logging
 import random
 from typing import Any, Iterator, Tuple
@@ -14,6 +15,22 @@ import tensorflow as tf
 from usl_models.flood_ml import constants
 from usl_models.flood_ml import metastore
 from usl_models.flood_ml import model
+
+
+@dataclasses.dataclass
+class Filters:
+    """Dataset filter settings."""
+
+    # Minimum mean value of a label to apply for filtering out sparse data.
+    min_mean_value: float = 0.0
+    # Ratio of samples to skip below the min_mean_value threshold [0.0, 1.0].
+    min_mean_value_dropout_rate: float = 0.0
+
+    def filter(self, label: tf.Tensor) -> bool:
+        """Returns true if the sample should be filtered."""
+        return (tf.reduce_mean(label) < self.min_mean_value) and (
+            random.random() < self.min_mean_value_dropout_rate
+        )
 
 
 def geospatial_dataset_signature() -> tf.TensorSpec:
@@ -55,6 +72,7 @@ def load_dataset(
     max_chunks: int | None = None,
     firestore_client: firestore.Client = None,
     storage_client: storage.Client | None = None,
+    filters: Filters | None = None,
 ) -> tf.data.Dataset:
     """Creates a dataset which generates chunks for the flood model.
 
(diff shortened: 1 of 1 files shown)


tatianawu:
I think the original idea that Walt proposed was adding a `percent_no_data` field to the metastore to actually calculate the % of each chunk that isn't semantically meaningful. I'm worried that filtering based on mean label will exclude chunks where we have small amounts of flooding that the model should learn on.

Katsutoshii:
> I think the original idea that Walt proposed was adding a `percent_no_data` field to the metastore to actually calculate the % of each chunk that isn't semantically meaningful. I'm worried that filtering based on mean label will exclude chunks where we have small amounts of flooding that the model should learn on.
> 
> IIRC there's a binary mask for the NODATA cells from the DEM. It should just be a mean over this mask, and then we can add another filter in `load_dataset` to only include chunks that exceed some threshold of available data.

Is this data already available in metastore? If not, does this mean-based approach still make sense to include as stopgap? With a sufficiently low threshold we shouldn't have any issue with over filtering.

tatianawu:
> > I think the original idea that Walt proposed was adding a `percent_no_data` field to the metastore to actually calculate the % of each chunk that isn't semantically meaningful. I'm worried that filtering based on mean label will exclude chunks where we have small amounts of flooding that the model should learn on.
> > IIRC there's a binary mask for the NODATA cells from the DEM. It should just be a mean over this mask, and then we can add another filter in `load_dataset` to only include chunks that exceed some threshold of available data.
> 
> Is this data already available in metastore? If not, does this mean-based approach still make sense to include as stopgap? With a sufficiently low threshold we shouldn't have any issue with over filtering.

I'm not sure; let's wait for @waltaskew to chime in once he's back. But in general, I feel like it makes more sense to put this information in the metastore rather than computing a mean each time we load in a tensor. I'd rather avoid introducing temporary code, especially since this isn't really anything urgent.

waltaskew:
I'd prefer not to leave folks with stopgaps for anything that isn't urgent (e.g. some stopgaps allowing them to make predictions are okay since that's a core thing they need to be able to do.) USL will need to be able to implement these sorts of experiments on their own starting Thursday! So I think it's more valuable to leave them with good patterns & good documentation they can follow rather than leaving them with stopgaps which allow us to finish tasks before Thursday.

Some options:
- putting together a notebook which demonstrates some improvements from removing empty chunks
- writing documentation describing this as future work with some pointers on how to implement it
- write the code we'd really want for this which calculates percent no-data in the data pipeline and re-run the pipeline to get those values into the metastore (this is a pretty easy task we could accomplish in a day, but I'd just weight the importance of running that experiment for them versus making sure they're set up to run future experiments themselves.)

tatianawu:
> I'd prefer not to leave folks with stopgaps for anything that isn't urgent (e.g. some stopgaps allowing them to make predictions are okay since that's a core thing they need to be able to do.) USL will need to be able to implement these sorts of experiments on their own starting Thursday! So I think it's more valuable to leave them with good patterns & good documentation they can follow rather than leaving them with stopgaps which allow us to finish tasks before Thursday.
> 
> Some options:
> 
> * put together a notebook which demonstrates some improvements from removing empty chunks
> * write documentation describing this as future work with some pointers on how to implement it
> * write the code we'd really want for this which calculates percent no-data in the data pipeline and re-run the pipeline to get those values into the metastore (this is a pretty easy task we could accomplish in a day, but I'd just weight the importance of running that experiment for them versus making sure they're set up to run future experiments themselves.)

I briefly touched upon [this idea](https://www.notion.so/climate-iq/ML-Iterations-After-NYC-Prototypes-fa77e4c3db834b8a9a9c76f8aa3d90d7?pvs=4#a300cdeed88e488d8d3985b86b71964c) when discussing iterating on the NYC prototype. We can add more detail to get USL folks set up better, but I think this might actually be a good initial exercise to get folks to work with multiple parts of the pipeline.

PR #160, opened 2024-07-30
Evaluation updates to FloodML

tatianawu:
* Updated metrics to compute over individual examples instead of batches
* Batch evaluation is still supported, but batch dimensions are included in the metric computations, which can aggregate away meaningful information
* Added a function for spatial NSE, which is a common time series metric in hydrodynamic modeling

### DIFF
--- a/usl_models/usl_models/flood_ml/eval.py
+++ b/usl_models/usl_models/flood_ml/eval.py
@@ -1,58 +1,91 @@
-"""Evaluation metrics for FloodML."""
+"""Evaluation metrics for FloodML.
+
+By default, all the calculations are assumed to be over a single prediction/label
+due to the computational cost of model prediction. If you want to calculate
+metrics *per prediction* over a batch, they should be unstacked into individual
+examples:
+
+```
+for predictions, labels in batch_predict():
+    for prediction, label in zip(tf.unstack(predictions), tf.unstack(labels)):
+        # calculate metrics
+```
+
+Otherwise, specifying `batch = True` in any of the metric functions will average
+over the examples.
+"""
 
 import tensorflow as tf
 
 
-def spatial_mae(predictions: tf.Tensor, labels: tf.Tensor) -> tf.Tensor:
-    """Calculates spatial MAE over a batch of predictions.
+def spatial_mae(
+    predictions: tf.Tensor, labels: tf.Tensor, batch: bool = False
+) -> tf.Tensor:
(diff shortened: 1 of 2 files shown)


SallyElHajjar (approved):

usl_models/usl_models/flood_ml/eval.py
```diff
@@ -1,58 +1,117 @@
...
+over the examples.
+"""
 
 import tensorflow as tf
 
 
-def spatial_mae(predictions: tf.Tensor, labels: tf.Tensor) -> tf.Tensor:
-    """Calculates spatial MAE over a batch of predictions.
+def spatial_mae(
+    predictions: tf.Tensor, labels: tf.Tensor, batch: bool = False
```
waltaskew, line 22: do you think it'd be too cute for the code itself to figure out if it is looking at a batch or not based on the shape of the tensors, rather than expecting callers to set a `batch` flag?

waltaskew (approved):

usl_models/usl_models/flood_ml/eval.py
tatianawu: It could! One concern might be that the user passes in an incorrect tensor with the right dimensions, e.g., if they try to pass in a single sequence of predictions `[T, H, W]` here, instead of a batch `[B, H, W]`. Though I guess there's nothing stopping them even now from that (we could add an `ensure_shape` as a validation step).

tatianawu: I also considered splitting up the logic into separate functions, like having a separate `batch_spatial_mae`. Do you have a preference on that versus the flag?

I'll go and add an `ensure_shape`.  It won't completely remove the possibility of failure cases, like mixing up batch and temporal axes, but it should catch most other bad inputs.

waltaskew: I have a slight preference for a separate `batch_spatial_mae` but I'm down for anything here.
I generally prefer to avoid dispatching behaviour on values when it can be avoided, like
```
do_x():

do_y():
```
vs
```
do_x_or_y(do_x: bool):
```
but again think it's all good in this PR and have approved it.

tatianawu: Separated the logic between single and batch predictions.

I realized that `ensure_shape` doesn't really work here because we don't know the specific shape; I was confusing it with rank. But added a couple of asserts for what I think we do know.

PR #161, opened 2024-07-31
Avoid division by zero in NSE calculation

tatianawu:


waltaskew (approved):
Would be great to add a test exercising this

tatianawu:
> Would be great to add a test exercising this

Do you mean adding a test that checks the behavior when we do divide by zero? It doesn't fail, but the tensor will include NaN and inf values.

waltaskew:
> > Would be great to add a test exercising this
> 
> Do you mean adding a test that checks the behavior when we do divide by zero? It doesn't fail, but the tensor will include NaN and inf values.

A mean a test which demonstrates us successfully evading a situation in which we would otherwise divide by zero and the result wouldn't include NaNs or infs

tatianawu:
> > > Would be great to add a test exercising this
> > 
> > 
> > Do you mean adding a test that checks the behavior when we do divide by zero? It doesn't fail, but the tensor will include NaN and inf values.
> 
> I mean a test which demonstrates us successfully evading a situation in which we would otherwise divide by zero and the result wouldn't include NaNs or infs

Done.

PR #162, opened 2024-08-12
Flood model improvement

SallyElHajjar:
I have added some improvements to the existing flood ML model.

### DIFF
--- a/usl_models/usl_models/flood_ml/model.py
+++ b/usl_models/usl_models/flood_ml/model.py
@@ -251,13 +251,16 @@ class FloodConvLSTM(tf.keras.Model):
                 # Remaining layers are TimeDistributed and are applied to each
                 # temporal slice
                 layers.TimeDistributed(layers.Conv2D(8, 5, **st_cnn_params)),
+                layers.TimeDistributed(layers.BatchNormalization()),  # Batch Norm
                 layers.TimeDistributed(
                     layers.MaxPool2D(pool_size=2, strides=1, padding="same")
                 ),
                 layers.TimeDistributed(layers.Conv2D(16, 5, **st_cnn_params)),
+                layers.TimeDistributed(layers.BatchNormalization()),  # Batch Norm
                 layers.TimeDistributed(
                     layers.MaxPool2D(pool_size=2, strides=1, padding="same")
                 ),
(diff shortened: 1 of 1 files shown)


usl_models/usl_models/flood_ml/model.py
```diff
@@ -251,13 +251,16 @@ def __init__(
                 # Remaining layers are TimeDistributed and are applied to each
                 # temporal slice
                 layers.TimeDistributed(layers.Conv2D(8, 5, **st_cnn_params)),
+                layers.TimeDistributed(layers.BatchNormalization()),  # Batch Norm
```
tatianawu, line 254: Like I mentioned before, we don't use BatchNorm because the model is currently constrained to a batch size of 4 during training. During inference, we're often just predicting on 1-2 examples (depending on hardware). This is probably just going to add noise unless you can increase the batch size.

Is this a viable experiment? Yes. Is it what I think should be the priority right now? No.

IMO fixing data sparsity should take precedence over any other model work. Otherwise the model is going to continue producing zeros.

SallyElHajjar: it is evident that we should do other enhancements over the data pipeline too, i am not saying that this is just the only solution that we are going to proceed with. Regarding the batch norm, i had a plan to test also LayerNorm or GroupNorm, i believe that it's possible with proper tuning, it could still contribute to the stability and generalization of the model even with small batch size.

tatianawu: You can and should test this locally before checking it in then. Remember that you can train with local changes, so there's no reason to check in all the changes you're making for preliminary experiments.

I don't necessarily agree that "it's evident" you need to make model "improvements" before you've fixed data sparsity, because there's no telling what other performance issues may be present. Yes, there's very likely going to be room for improvement, but I don't see why that translates into immediately playing around with BatchNorm instead of doing things like longer horizon forecasting.

SallyElHajjar: i am saying it is evident for the data pipeline which includes the data sparsity handling, not the model improvement.

Katsutoshii: +1 to Tatiana's points. In general, I think before checking in model changes we should have data suggesting that this improves model performance.

PR #165, opened 2024-08-22
LANDUSE_INDEX normalization

SallyElHajjar:
I have added the LANDUSE_INDEX normalization & testing files. I have also added the testing files. 
I have also added "handling the case where we have missing features"


Katsutoshii (approved):

usl_models/tests/atmo_ml/atmo_model_test.py
```diff
@@ -55,10 +59,16 @@ def fake_input_batch(
...
+    LU_INDEX = tf.random.uniform(
+        (batch_size, height, width),
+        minval=0,
+        maxval=_LU_INDEX_VOCAB_SIZE,
+        dtype=tf.int32,
+    )
     return {
         "spatial": spatial,
         "spatiotemporal": spatiotemporal,
+        "LU_INDEX": LU_INDEX,
```
Katsutoshii, line 71: For consistency can we make this lowercase?

SallyElHajjar: was thinking to write it in this way to align with the data pipeline. (the exact feature is defined in capslock) is that necessarily? or I can put it here in lowercase?

PR #166, opened 2024-08-26
Visualization code and dependency updates

Katsutoshii:
I've added logic for running data visualization in colabs to usl_models lib.
Also to streamline installation of the package on new dev environments, I've synced requirements.txt with the contents of setup.py.


SallyElHajjar (approved):
LGTM!!

PR #167, opened 2024-08-28
adding land ue index to pipeline

SallyElHajjar:
I have added the land use index to the data pipeline. I have added it to the end to the file, as i remember it was required to do so.
 I have added it at the end, as it was required by Kathryn, she has already tested the feature/ label matrices.

### DIFF
--- a/usl_pipeline/usl_lib/usl_lib/shared/wps_data.py
+++ b/usl_pipeline/usl_lib/usl_lib/shared/wps_data.py
@@ -320,5 +320,12 @@ ML_REQUIRED_VARS_REPO = dict(
                 "type": ScalingType.NONE,
             },
         },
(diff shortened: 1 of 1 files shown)


MadhaviJain90:
Pls refer to wps_vars branch in the repo. I would strongly suggest to move LU_INDEX, below LANDUSEF. Otherwise it would create confusion as more variables are added or when some need to be deleted in the future. https://github.com/UrbanSystemsLab/climateiq-cnn/commit/50042fbdcb9ef2cff249fd1f582f67947f282d97#diff-5349d7e7eb77dc7925f832b58222c20817b96caa0cabc411a395f0af1c55f60b)

MadhaviJain90 (approved):
LGTM!!!

PR #169, opened 2024-09-11
spatial Window

SallyElHajjar:
@Katsutoshii I have removed the stack that was at the end, and also squeezing.  I have also corrected the inputs/outputs


usl_models/usl_models/atmo_ml/input_output_sequences.py
```diff
@@ -0,0 +1,117 @@
...
+                    labels[:, day * time_steps_per_day + t],  # Y^t
+                    labels[:, day * time_steps_per_day + (t + 1)],  # Y^{t+1}
+                ],
+                axis=-1,
+            )
+            daily_output_sequences.append(output_seq)
+
+        # Instead of stacking, we process sequences directly (e.g., pass to model)
+        for input_seq, output_seq in zip(daily_input_sequences, daily_output_sequences):
+            tf.print("Input sequence:", input_seq.numpy().flatten().tolist())
```
Katsutoshii, line 115: Let's guard these print statements behind a `debug: bool` parameter in the function. That way we won't get log spam when not debugging.

Katsutoshii (approved):

usl_models/usl_models/atmo_ml/input_output_sequences.py
SallyElHajjar: Done

PR #172, opened 2024-09-16
Add HTTP triggerable functions for testing data pipeline

Katsutoshii:
Add HTTP triggerable functions for testing data pipeline

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -169,12 +169,13 @@ def build_and_upload_study_area_chunk(
 @functions_framework.http
 def build_and_upload_study_area_chunk_http(
     request: flask.Request,
-) -> None:
+) -> flask.Response:
     request_json = request.get_json()
-    return _build_and_upload_study_area_chunk(
+    _build_and_upload_study_area_chunk(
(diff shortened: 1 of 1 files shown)


SallyElHajjar:
> Add HTTP triggerable functions for testing data pipeline
 Regarding the checks:
 The issue stems from the fact that the functions you have defined for the HTTP endpoints (build_and_upload_study_area_chunk_http, build_wrf_label_matrix_http, and build_feature_matrix_http) currently return None, but mypy expects them to return a specific response type, such as Response, str, bytes, etc.
Solution:
For Flask-based HTTP functions, we need to return a valid HTTP response, such as flask.Response, a string, or a JSON object. 
maybe doing the following could help? and same for other functions?
build_and_upload_study_area_chunk_http:

@functions_framework.http
def build_and_upload_study_area_chunk_http(
    request: flask.Request,
) -> flask.Response:
    request_json = request.get_json()
    _build_and_upload_study_area_chunk(
        file_name=pathlib.PurePosixPath(request_json["name"]),
        bucket_name=request_json["bucket"],
    )
    return flask.jsonify({"status": "Study area chunk built and uploaded."})

SallyElHajjar (approved):

PR #170, opened 2024-09-12
added solar time sin and cos to main.py

MadhaviJain90:

PR #174, opened 2024-09-17
added the solar time

SallyElHajjar:
I have added the solar time and tested it in my local.  I have changed the expected values to remove the errors. Could you confirm ? @MadhaviJain90

### DIFF
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -1037,12 +1037,19 @@ def _compute_solar_time_components(dataset: xarray.Dataset) -> xarray.Dataset:
         latitudes = dataset["XLAT_M"][0, :, :]
         times = dataset["Times"]
 
-        # Convert time variable to datetime objects (assume Times is in seconds)
-        # Modify the time unit as per your dataset's definition
-        time_unit = "seconds since 1970-01-01 00:00:00"
+        # Convert time variable to datetime objects
         times = xarray.DataArray(
             [
-                numpy.datetime64(netCDF4.num2date(t, time_unit).isoformat())
+                # Check if the value is a Unix timestamp (numeric type)
+                (
+                    numpy.datetime64(
+                        datetime.datetime.utcfromtimestamp(t).strftime(
+                            "%Y-%m-%dT%H:%M:%S"
+                        )
+                    )
+                    if isinstance(t, (int, float))
+                    else numpy.datetime64("".join(t.astype(str)).replace("_", "T"))
+                )
                 for t in times.values
             ]
         )
--- a/usl_pipeline/cloud_functions/main.py
+++ b/usl_pipeline/cloud_functions/main.py
@@ -1054,15 +1054,21 @@ def _compute_solar_time_components(dataset: xarray.Dataset) -> xarray.Dataset:
             ]
         )
 
-        # Extract hours and minutes from the datetime objects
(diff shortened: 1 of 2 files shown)


usl_pipeline/cloud_functions/main.py
```diff
@@ -1027,6 +1028,74 @@ def _direction_to_cosine(degrees):
     return dataset
 
 
+def _compute_solar_time_components(dataset: xarray.Dataset) -> xarray.Dataset:
+    # Solar Time Computation for every Longitude of the City
+    # Solar Time Sine and Cosine Derivation from there
+    if all(var in dataset.keys() for var in ["XLONG_M", "XLAT_M", "Times"]):
```
Katsutoshii, line 1034: How should we handle the case when these features are missing? Should we report an error to the user?

Katsutoshii (approved):

usl_pipeline/cloud_functions/main.py
```diff
@@ -1027,6 +1028,81 @@ def _direction_to_cosine(degrees):
...
+
+        # Extract hours and minutes from the datetime objects
+        utc_hours_minutes = (
+            times.astype("datetime64[h]").astype(int) % 24
+            + times.astype("datetime64[m]").astype(int) % 60 / 60
+        )
+
+        # Define the function to calculate solar time
+        def calculate_solar_time(utc_time, longitude):
+            return (utc_time + longitude / 15) % 24
```
ahmustafa, line 1065: what if we have negative values of longitude (west of the prime meridian)?
we may adjust as 
return ((utc_time + longitude / 15) % 24 + 24) % 24

MadhaviJain90: For negative longitudes like NYC the code will take care of it. Eg, if WRF variable XLONG_M = -75, then local time would be UTC-(5) hrs. What I am concerned about is the sin and cos solar time values that I was getting in my code and the values that the unit tests in this one is giving

Katsutoshii (approved):

MadhaviJain90:
I don't see any differences from the Solar Time PR I had coded -> https://github.com/UrbanSystemsLab/climateiq-cnn/pull/170, except a minor alteration in the expected value of sin and cos of solar time. LGTM. Can be merged!

MadhaviJain90 (approved):
can be merged

SallyElHajjar:
@MadhaviJain90   
 The difference is that time was handled incorrectly in your previous PR, (error when testing). This is the mainly edited part:
         # Convert time variable to datetime objects
        times = xarray.DataArray(
            [
                # Check if the value is a Unix timestamp (numeric type)
                (
                    numpy.datetime64(
                        datetime.datetime.utcfromtimestamp(t).strftime(
                            "%Y-%m-%dT%H:%M:%S"
                        )
                    )
                    if isinstance(t, (int, float))
                    else numpy.datetime64("".join(t.astype(str)).replace("_", "T"))
                )
                for t in times.values
            ]
        )

        print("Debug - Converted Times:", times)

        # Extract hours and minutes from the datetime objects separately
        utc_hours = times.dt.hour  # Extract the hour component
        utc_minutes = times.dt.minute  # Extract the minute component

        # Convert to fractional hours (hours + minutes/60)
        utc_hours_minutes = utc_hours + utc_minutes / 60.0

        print("Debug - UTC Hours and Minutes (in fraction):", utc_hours_minutes)

PR #176, opened 2024-09-18
correcting the  in def create_input_output_sequences

SallyElHajjar:
I have created the create_input_output_sequences function and added a debug instruction.


usl_models/usl_models/atmo_ml/input_output_sequences.py
```diff
@@ -5,7 +5,7 @@
 
 
 def create_input_output_sequences(
-    inputs, labels, time_steps_per_day=constants.TOTAL_TIME_STEPS, debug=False
+    inputs, labels, time_steps_per_day=constants.INPUT_TIME_STEPS, debug=True
```
Katsutoshii, line 8: Let's leave this default to false.

SallyElHajjar: done

PR #181, opened 2024-10-04
Solarsally

SallyElHajjar:
I have added two options for the testing file, one for GCP and another for fake data. I have also tested both of them locally, it works with no error.  Now the same code handles both fake and real data. @Katsutoshii


Katsutoshii (approved):
LGTM, just move the remote data dependency to a separate python file so it's not under unit test.

usl_pipeline/cloud_functions/main_test.py
```diff
@@ -652,59 +652,40 @@ def test_compute_wind_components():
...
-
-    # Close the NetCDF file and retrieve its contents as bytes
-    memfile = ncfile.close()
-    ncfile_bytes = memfile.tobytes()
-
-    # Open the in-memory NetCDF file with xarray
-    ds = xarray.open_dataset(io.BytesIO(ncfile_bytes))
-
+    # Initialize Google Cloud Storage client
+    storage_client = storage.Client()
```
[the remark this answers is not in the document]
SallyElHajjar: done

usl_pipeline/cloud_functions/main_test.py
```diff
@@ -651,60 +651,93 @@ def test_compute_wind_components():
...
+            [[-74.00, -73.90], [-74.10, -73.80]], dtype=numpy.float32
+        )
+        latitudes[:] = numpy.array(
+            [[40.70, 40.70], [40.80, 40.80]], dtype=numpy.float32
+        )
 
-    # Set the time using numpy.datetime64 for the expected format
-    times[0] = numpy.datetime64("2010-02-02T18:00:00")
+        # Close the NetCDF file and retrieve its contents as bytes
+        memfile = ncfile.close()
```
ahmustafa: based on my understanding we retrieve the contents of ncfile using memfile.tobytes(). I'm concerned about memfile = ncfile.close(), because the close() method doesn’t return any value, leading to a potential error.
just be careful and consider not closing the NetCDF file until after assigned the necessary data for xarray

SallyElHajjar: This case is with in-memory NetCDF files using netCDF4.Dataset, the behavior is a bit different than in typical file handling when the memory parameter is set to 1 (indicating in-memory storage). By specifying memory=1, we are instructing netCDF4 to create a file in memory instead of on disk, which changes the behavior of how the file is handled. now when we use "memfile = ncfile.close()" , instead of simply closing the file, the in-memory buffer (which contains the NetCDF file's contents) is returned. This buffer is similar to a file-like object, and we can then access the contents in bytes.

PR #183, opened 2024-10-14
Use ubuntu-22.04 instead of ubuntu-latest in GitHub workflows

SallyElHajjar:
Use ubuntu-22.04 instead of ubuntu-latest in GitHub workflows


Katsutoshii (approved):
Please rename the PR to "Use ubuntu-22.04 instead of ubuntu-latest in GitHub workflows"

PR #186, opened 2024-10-18
Update wps_data to use an enum for WPS variable.

Katsutoshii:
Update wps_data to use an enum for WPS variable.
This allows us to use the index value in ML dataset loader.


SallyElHajjar:
For checks:
**1. No overload variant of "__setitem__" (Line 984):**
This error suggests that you're trying to assign an ndarray to a list at a specific index, but the types don't match. Python's built-in list supports only a limited set of types for the __setitem__ method.
**Solution:** Ensure that both the list and the element you're assigning to it have compatible types.
If the list should store ndarray objects, make sure it's initialized with the correct type, like this:
```
from typing import List
import numpy as np

my_list: List[np.ndarray] = []
my_list[index] = np.array([1, 2, 3])  # Example assignment

```
Alternatively, convert the ndarray to a type that the list can handle, if necessary:
`my_list[index] = my_ndarray.tolist()  # Convert ndarray to list`
**2. Need type annotation for "features_matrix" (Line 986):**
This error occurs because mypy requires you to explicitly annotate the type of the features_matrix variable.

**Solution:** Add the type annotation based on what kind of data features_matrix is expected to hold. For example:
```
import numpy as np
from typing import List

features_matrix: List[np.ndarray] = []
```
**If it’s an array:**
`features_matrix: np.ndarray = np.array([...])`
**3. Incompatible argument for dstack (Line 986):**
The dstack function from NumPy expects a sequence of arrays, but it seems like you're passing a list that contains None.

**Solution:** Ensure that the list you pass to dstack contains valid arrays. If None values are present in the list, filter them out:
```
import numpy as np

valid_arrays = [arr for arr in features_matrix if arr is not None]
result = np.dstack(valid_arrays)

```
Make sure that the features_matrix is a list or sequence of arrays, not a list of None values.

SallyElHajjar (approved):

ahmustafa (approved):

PR #190, opened 2024-10-25
Revert "Rename Cloud Functions CI"

jainrajan98:
We shouldn't do merge before updating the branch protection rule to fix the typo


SallyElHajjar:
@Katsutoshii i think yes from my previous discussion with Joisah, we shouldnt merge it right? @Katsutoshii @jainrajan98

SallyElHajjar (approved):

PR #198, opened 2024-12-12
testing job checks

SallyElHajjar:
testing job checks

### DIFF
--- a/usl_models/tests/atmo_ml/atmo_model_test.py
+++ b/usl_models/tests/atmo_ml/atmo_model_test.py
@@ -13,6 +13,7 @@ _TEST_SPATIAL_FEATURES = 22  # lu_index is now separate
 _TEST_SPATIOTEMPORAL_FEATURES = 12
 _LU_INDEX_VOCAB_SIZE = 61
 _EMBEDDING_DIM = 8
+# testing job checks
 
 
 def pytest_model_params() -> model_params.AtmoModelParams:


usl_models/notebooks/train_atmo_model copy.ipynb
Katsutoshii, line 1: Let's remove the _copy notebooks unless they are generally useful for future work. If we do keep them, give them a more meaningful name.

usl_models/tests/atmo_ml/dataset_test.py
Katsutoshii, line 1: Instead of commenting out the entire file, can you put an early return on each of the failing tests?

usl_models/usl_models/atmo_ml/constants.py
```diff
@@ -1,12 +1,13 @@
 """Constant definitions for the AtmoML model."""
 
-# Geospatial constants
-MAP_HEIGHT = 200
-MAP_WIDTH = 200
+# Geospatial constants should be 200
```
Katsutoshii, line 3: Should we check this in? I thought this was for experimental purposes only.

usl_models/training_log.txt
Katsutoshii, line 1: We should add this to the gitignore.

Katsutoshii (changes requested):
At a high level, a lot of this seems like experimental code that shouldn't be checked in. Can we start by checking in the dataset changes only?

usl_models/usl_models/atmo_ml/constants.py
SallyElHajjar: was thinking about that, but thought if I keep 200 with the new model architecture i will get the memory error again because the model will not be built. I can try do that

usl_models/notebooks/train_atmo_model copy.ipynb
SallyElHajjar: okkay

usl_models/training_log.txt
SallyElHajjar: ohh i forgot to delete it. will remove it because we are no longer using it

PR #199, opened 2025-01-23
Implement bulk file downloader to be used for caching.

Katsutoshii:
Implement bulk file downloader to be used for caching.


SallyElHajjar:
@Katsutoshii will just have to correct the flkake8 error before merging

SallyElHajjar (approved):

PR #207, opened 2025-02-19
Evaluation metrics and visualization

SallyElHajjar:
Adding some evaluation metrics and adding the plot of difference between the predicted map and Ground truth.

### DIFF
--- a/usl_models/usl_models/atmo_ml/metrics.py
+++ b/usl_models/usl_models/atmo_ml/metrics.py
@@ -31,3 +31,101 @@ class OutputVarMeanSquaredError(keras.metrics.MeanMetricWrapper):
     def get_config(self):
         """Returns the serializable config of the metric."""
         return {"name": self.name, "dtype": self.dtype, "sto_var": self.sto_var}
+
+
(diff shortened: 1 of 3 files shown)


usl_models/usl_models/atmo_ml/metrics.py
```diff
@@ -31,3 +31,142 @@ def mse(y_true: tf.Tensor, y_pred: tf.Tensor):
     def get_config(self):
         """Returns the serializable config of the metric."""
         return {"name": self.name, "dtype": self.dtype, "sto_var": self.sto_var}
+
+
+@keras.saving.register_keras_serializable(package="CustomMetrics")
+class SSIMMetric(keras.metrics.Metric):
```
Katsutoshii, line 37: Why do these metrics need to be stateful? Couldn't they be stateless?

Katsutoshii: Ah I guess it's to compute it per example not per batch. In that case, can we have some unit tests to ensure these work as intended?

Katsutoshii (changes requested):
Wh

Katsutoshii (approved):

PR #211, opened 2025-02-22
AtmoModel.Params dataclass

Katsutoshii:
Migrate AtmoModel.Params to a dataclass, which is easier to have defaults with.


SallyElHajjar:
@Katsutoshii there are some flake8 errrors

SallyElHajjar (approved):

PR #213, opened 2025-02-26
Learning rate

SallyElHajjar:
# Use an exponential decay schedule for a smoother learning rate reduction.

### DIFF
--- a/usl_models/usl_models/atmo_ml/model.py
+++ b/usl_models/usl_models/atmo_ml/model.py
@@ -153,9 +153,17 @@ class AtmoModel:
 
     def _build_model(self) -> keras.Model:
         """Creates the correct internal (Keras) model architecture."""
+         # Use an exponential decay schedule for a smoother learning rate reduction.
+        lr_schedule = tf.keras.optimizers.schedules.ExponentialDecay(
+            initial_learning_rate=1e-3,
+            decay_steps=1000,
+            decay_rate=0.9,
+            staircase=True,
+        )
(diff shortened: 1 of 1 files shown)


usl_models/usl_models/atmo_ml/model.py
```diff
@@ -153,9 +153,17 @@ def from_checkpoint(cls, artifact_uri: str, **kwargs) -> "AtmoModel":
 
     def _build_model(self) -> keras.Model:
         """Creates the correct internal (Keras) model architecture."""
+        # Use an exponential decay schedule for a smoother learning rate reduction.
+        lr_schedule = tf.keras.optimizers.schedules.ExponentialDecay(
```
Katsutoshii, line 157: Can you move this under the default value for the model params?

Katsutoshii:
Can you show a tensorboard plot comparing the training run before and after this change?

usl_models/usl_models/atmo_ml/model.py
SallyElHajjar: Done

SallyElHajjar:
WITHOUTDECAY

Katsutoshii:
What's the change in MSE @1000 epochs?

SallyElHajjar:
@Katsutoshii wasn't able to get the plots because of tensorboard strange issue. Didn't found the plot after I made the change (it looks like one point. ThAT's why I have just reported these plots. Is that necessarily to repeat or these plots work?

Katsutoshii:
Let's repeat the run, I'd like to have a loss number we can compare against previous runs.

PR #214, opened 2025-02-28
Separate model param input_cnn_kernel_size.

Katsutoshii:
Separate model param input_cnn_kernel_size.


usl_models/usl_models/atmo_ml/model.py
```diff
@@ -29,7 +29,10 @@ class AtmoModel:
     class Params(keras_dataclasses.Base):
         """Model parameters."""
 
-        # Layer-specific parameters.
+        # Input CNN params
+        input_cnn_kernel_size: int = 5
```
SallyElHajjar, line 33: should we also make it different for different CNN blocks?

PR #219, opened 2025-03-08
Drop WindDir Sin/Cos

Katsutoshii:
Drop WindDir Sin/Cos


Katsutoshii:
Screenshot From 2025-03-07 17-57-33
MSE T2 @1000 epochs 1.33e-4

Screenshot From 2025-03-07 18-00-43
MSE RH @1000 epochs 6.86e-4

PR #220, opened 2025-03-08
Implement dense decoder instead of ConvTranspose

Katsutoshii:
Implement dense decoder instead of ConvTranspose.


Katsutoshii:
Screenshot From 2025-03-07 18-37-52

MSE@1000 at 3.02e-4.

Qualitatively, nicer looking model results:
Screenshot From 2025-03-07 18-38-42

PR #221, opened 2025-03-11
Update the visualizer

SallyElHajjar:
Revert the normalization and add separated Vmin and Vmax for GT and prediction's plots.


Katsutoshii:
Why do we want a different scale for ground truth versus predictions? Isn't it easier to compare with the same scale?

SallyElHajjar:
I have added the dynamic option, But also kept Vmin and Vmax. Madhavi mentioned that she needs it to explicitly put some ranges to check the data. Now the duynamic option if set to true, it automatically plots the values of the data, if False, ours. @Katsutoshii

Katsutoshii:
I get the use case for passing a vmin and vmax, but I don't think we should support separate ranges for ground truth versus prediction. If the ranges need to be different then we know the model isn't working well anyways, so clamping values at the custom range should be fine.

usl_models/usl_models/atmo_ml/visualizer.py
```diff
@@ -24,11 +36,17 @@ def plot_2d_timeseries(
...
+    dynamic_colorscale: bool = False,
 ) -> matplotlib.figure.Figure:
-    """Plot a timeseries of 2D maps without normalization if desired."""
+    """Plot a timeseries of 2D maps.
+
+    If dynamic_colorscale is True, compute vmin and vmax from the data.
+    """
     T, H, W, *_ = data.shape
 
+    if dynamic_colorscale:
```
Katsutoshii, line 47: Let's compute this over both pred and label?

Katsutoshii (changes requested):

usl_models/usl_models/atmo_ml/visualizer.py
SallyElHajjar: done

jainrajan98 (approved):

jainrajan98 (approved):

PR #223, opened 2025-03-12
Use reflection padding instead of constant padding.

Katsutoshii:
Use reflection padding instead of constant padding.


Katsutoshii:
Screenshot From 2025-03-12 17-00-07
MSE@1000 = 0.0040 from 0.0051.

usl_models/usl_models/shared/pad_layers.py
```diff
@@ -0,0 +1,37 @@
...
+        spatial dimensions.
+        mode: Padding mode.
+
+    Returns:
+        A padded tensor with the same type as the input tensor.
+    """
+
+    def __init__(
+        self, padding: Tuple[int, int] = (1, 1), mode: str = "REFLECT", **kwargs
+    ):
```
SallyElHajjar, line 20: should we consider making the padding values and mode parameters part of the model parameters (if they need to be configurable) rather than hard-coding them in the layer? or not necessarily? @Katsutoshii

usl_models/usl_models/shared/pad_layers.py
```diff
@@ -0,0 +1,37 @@
...
+        """Constructor."""
+        self.padding = padding
+        self.mode = mode
+        super(Pad2D, self).__init__(**kwargs)
+
+    def compute_output_shape(self, shape):
+        """Compute output shape."""
+        h_pad, w_pad = self.padding
+        B, H, W, *S = shape
+        return (B, H + 2 * h_pad, W + 2 * w_pad, *S)
```
SallyElHajjar, line 30: we can also do somethibng like new_h = input_shape[1] + (h_pad * 2) if input_shape[1] is not None else None; same for new_w in case we  expect variable input shapes during inference or training. Dont think will do that, but what are your thoughts?

SallyElHajjar (approved):

usl_models/usl_models/shared/pad_layers.py
Katsutoshii: Good call, will add to AtmoModel.Params

usl_models/usl_models/shared/pad_layers.py
Katsutoshii: Turns out we don't need this implementation at all, removed.

PR #224, opened 2025-03-14
Use Conv layers instead of TransConv layers

Katsutoshii:
Use Conv layers instead of TransConv layers


usl_models/usl_models/atmo_ml/model.py
```diff
@@ -434,43 +433,36 @@ def __init__(self, params: AtmoModel.Params):
 
         # Output CNNs (upsampling via TransposeConv)
         # We return separate sub-models (i.e., branches) for each output.
-        output_cnn_params = ConvParams(
-            padding="valid", activation=self._params.output_activation
-        )
+        output_cnn_params = ConvTransposeParams(activation="relu", padding="valid")
```
SallyElHajjar, line 436: should we rename the original class convParams instead of convtransposee since we are no longer transpose?

SallyElHajjar (changes requested):

usl_models/usl_models/atmo_ml/model.py
Katsutoshii: Done.

SallyElHajjar (approved):

PR #234, opened 2025-04-08
Attention and ConvLSTM layers

SallyElHajjar:
adding attention to time dimension and three convlstm layers

### DIFF
--- a/usl_models/usl_models/atmo_ml/model.py
+++ b/usl_models/usl_models/atmo_ml/model.py
@@ -423,6 +423,36 @@ class AtmoConvLSTM(keras.Model):
                     dropout=self._params.lstm_dropout,
                     recurrent_dropout=self._params.lstm_recurrent_dropout,
                 ),
+                # Input shape: (time_steps, height, width, channels)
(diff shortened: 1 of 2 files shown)


SallyElHajjar:
image

image

image


image


First two: Three layers (without attention vs With attention)
Second two: one layer without attention vs with

usl_models/usl_models/atmo_ml/model.py
```diff
@@ -423,6 +423,36 @@ def __init__(self, params: AtmoModel.Params):
...
+                    pad_layers.Pad2D((LSTM_K_PAD, LSTM_K_PAD), mode=PAD_MODE)
+                ),
+                layers.ConvLSTM2D(
+                    LSTM_FILTERS,
+                    LSTM_K_SIZE,
+                    return_sequences=True,
+                    strides=1,
+                    padding="valid",
+                    activation=self._params.lstm_activation,
+                    dropout=self._params.lstm_dropout,
```
ahmustafa, line 438: why do we have the exact same block of layers twice in a row?

